PR #283, opened 2015-06-11
Update mqtt_server.php

haixingdev:
修改contect ack的返回


matyhtf:
已迁移至 https://github.com/swoole/mqtt/blob/master/server.php

PR #294, opened 2015-06-17
there is a memory leak in response->header

Lynnworld:


matyhtf:
This bug has been fixed.
https://github.com/swoole/swoole-src/commit/a0f632fa14437c23b7b9f42f092e477ef5132457

PR #336, opened 2015-08-04
Fixes typo when introducing async io

netroby:


Atriedes:
could you please also help translate swoole docs to english?

netroby:
Yes, i would like to help. but not sure they need my help.

PR #365, opened 2015-09-09
Fixed repeated memory allocations for in_buffer.

bryceliu:
When reactor_num is greater than 1, in_buffer for the same worker's pipe_master is created more than once(one time per reactor thread), which results in memory leaking.


matyhtf:
tks.

PR #426, opened 2015-12-10
Update http2.go

zzjin:
use go fasthttp

### DIFF
--- a/examples/bench/http2.go
+++ b/examples/bench/http2.go
@@ -1,29 +1,27 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


matyhtf:
fasthttp不支持Http KeepAlive，无法使用长连接进行测试。

zzjin:
Currently fasthttp is successfully used in a production serving 100K rps from 1M concurrent keep-alive connections on a single server. 官方不是支持吗?我本地ab测试-k是可以的...

PR #810, opened 2016-08-12
Reset start status after server shutdown

Fishdrowned:
The start status which is set in `swServer_start()`:
```c
568    SwooleGS->start = 1;
```
Should be reset after shoutdown


Fishdrowned:
Fixed in e40132b

PR #1016, opened 2017-01-11
Fix possible file descriptor leak

rvantonder:
swoole_file_get_contents may return without closing an open file descriptor.


matyhtf:
Thanks!

PR #1031, opened 2017-01-20
support bind port 0 to let os choose avaiable port

lvht:
As micro service is becoming more and more popular, service's port and address has been registered into some registery like zookeeper, etcd. There is no strong need to listen on certain port. The more service we have, the more port we should manage.

This PR promotes to enable swoole to choose listen port by OS. What we should do is like this,
```php
<?php
$serv = new Swoole\Http\Server("127.0.0.1"); // port is optional now

$serv->on('Request', function($request, $response) {
    $response->end("<h1>Hello Swoole!</h1>");
});

$serv->on('managerStart', function (Swoole\Server $s) {
    var_dump($s->host);
    var_dump($s->port);
    // send host:port to etcd, zookeeper ...
});

$serv->start();
```

### DIFF
--- a/include/swoole.h
+++ b/include/swoole.h
@@ -1201,7 +1201,7 @@ static sw_inline uint64_t swoole_ntoh64(uint64_t net)
 }
 
 int swSocket_create(int type);
-int swSocket_bind(int sock, int type, char *host, int port);
+int swSocket_bind(int sock, int type, char *host, int *port);
 int swSocket_wait(int fd, int timeout_ms, int events);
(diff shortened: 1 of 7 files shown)


matyhtf:
Good job,
The `swoole_server->listen` method also needs to be implemented.

lvht:
@matyhtf if you want listen to port 0, you just need 
```php
<?php
$serv->listen('127.0.0.1', 0, $TYPE);

$serv->on('managerStart', function (Swoole\Server $s) {
    var_dump($s->ports);
    // send host:port to etcd, zookeeper ...
});
```

matyhtf:
Ok, i see.  Merged

PR #1041, opened 2017-01-31
Fix memory leak in swTableColumn_add()

tutanhamon:
Reference: Coverity CID #1397914

### DIFF
--- a/src/memory/Table.c
+++ b/src/memory/Table.c
@@ -86,6 +86,7 @@ int swTableColumn_add(swTable *table, char *name, int len, int type, int size)
     col->name = swString_dup(name, len);
     if (!col->name)
     {
+        swTableColumn_free(col);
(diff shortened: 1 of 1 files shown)


matyhtf:
I have commited a fix code.
https://github.com/swoole/swoole-src/commit/9658d28ee05b71ef0440014539fec8b828fb9681

tutanhamon:
There is still a memory leak in `default switch(type)` in [Table.c  @  line 129](https://github.com/swoole/swoole-src/blob/9658d28ee05b71ef0440014539fec8b828fb9681/src/memory/Table.c#L129)

matyhtf:
https://github.com/swoole/swoole-src/commit/e5907f2138f7b4f3fc464014c9b1abb46c3211d0

fixed

PR #1062, opened 2017-01-31
Fix buffer over-read in swoole_mkdir_recursive()

tutanhamon:
Reference: Coverity CID #1397858


matyhtf:
```c
tmp[PATH_MAX] = 0;
```
 memory over write 1 byte here.

tutanhamon:
BTW, does it have much sense to make a truncated directory name? It had better exit with error if the directory name is too long.

PR #1092, opened 2017-02-15
set content-length for empty body

lvht:


lvht:
You @matyhtf fixed this problem by 651be6e309d8130297b3923f9b9a4d34f254f897. However, I think the duplicated code should be clear as soon as possible. Just FYI.

PR #1097, opened 2017-02-17
move protocol to connection

lvht:
* support multiple protocols on the same port
* prepare for http/1.1 upgrade

### DIFF
--- a/include/swoole.h
+++ b/include/swoole.h
@@ -411,7 +411,31 @@ typedef struct
     socklen_t len;
 } swSocketAddress;
 
-typedef struct _swConnection
+typedef struct _swConnection swConnection;
+typedef struct _swEvent swEvent;
+typedef struct _swReactor swReactor;
+typedef struct _swProtocol swProtocol;
+
+struct _swProtocol
+{
+    /* one package: eof check */
+    uint8_t split_by_eof;
+    uint8_t package_eof_len;  //数据缓存结束符长度
+    char package_eof[SW_DATA_EOF_MAXLEN + 1];  //数据缓存结束符
+
+    char package_length_type;  //length field type
+    uint8_t package_length_size;
+    uint16_t package_length_offset;  //第几个字节开始表示长度
+    uint16_t package_body_offset;  //第几个字节开始计算长度
+    uint32_t package_max_length;
+
+    void *private_data;
+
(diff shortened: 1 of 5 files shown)


include/swoole.h
```diff
@@ -411,7 +411,31 @@ typedef struct
     socklen_t len;
 } swSocketAddress;
 
-typedef struct _swConnection
+typedef struct _swConnection swConnection;
+typedef struct _swEvent swEvent;
+typedef struct _swReactor swReactor;
+typedef struct _swProtocol swProtocol;
+
+struct _swProtocol
```
lvht: no changed so much. Only move `struct _swProtocol` ahead of `struct _swConnection`

include/swoole.h
```diff
@@ -555,26 +579,10 @@ typedef struct _swConnection
...
-
-    char package_length_type;  //length field type
-    uint8_t package_length_size;
-    uint16_t package_length_offset;  //第几个字节开始表示长度
-    uint16_t package_body_offset;  //第几个字节开始计算长度
-    uint32_t package_max_length;
-
-    void *private_data;
+    swProtocol protocol;
+    int (*onRead)(swReactor *reactor, struct _swConnection *conn, swEvent *event);
```
lvht: now, protocol and onRead function has been added to connection, which make it possible to support different protocol on the same port.

src/network/ReactorThread.c
```diff
@@ -852,7 +852,7 @@ static int swReactorThread_onRead(swReactor *reactor, swEvent *event)
     }
 #endif
     event->socket->last_time = SwooleGS->now;
-    return port->onRead(reactor, port, event);
+    return event->socket->onRead(reactor, event->socket, event);
```
lvht: Now, call `connection->onRead`

src/network/Server.c
```diff
@@ -1394,6 +1394,9 @@ static swConnection* swServer_connection_new(swServer *serv, swListenPort *ls, i
     connection = &(serv->connection_list[fd]);
     bzero(connection, sizeof(swConnection));
 
+    memcpy((void *)&connection->protocol, (const void *)&ls->protocol, sizeof(ls->protocol));
+    connection->onRead = ls->onRead;
+
```
lvht: copy protocol and onRead function for every new accepted connection.

PR #1106, opened 2017-03-01
support socket linger for client

lvht:
support linger to let client perform fast close and reduce the number of server time-wait socket


matyhtf:
单独设置此参数无意义，一般是在close时根据情况设置的。

PR #1132, opened 2017-03-10
add pthread_mutexattr_destroy()

jellor:


matyhtf:
Should be destroyed in the free phase.

PR #1199, opened 2017-04-25
Disallow swoole_server->finish be called in onPipeMessage callback.

AJSoft:
If swoole_server->finish is called in onPipeMessage callback, and the source process is a task worker, segfault may occurred.
see issue #1198


matyhtf:
对`src/network/ReactorThread.c`的修改需要删除，否则会导致UDP的IP不正确。底层目前已修复此问题。

AJSoft:
已经revert

PR #1361, opened 2017-09-27
Fix building with libressl.

vanillahsu:
I try to build swoole-1.9.21 on FreeBSD with libressl, but could not find some constants.

here is the patch for libressl, please help to check, thanks.


matyhtf:
Thanks

PR #1382, opened 2017-10-30
bugfix

kimkit:


matyhtf:
It is not correct, your code can only filter a blank character.

kimkit:
this method is used in `examples/test_server.c`, but the response is not right.
the old one always replace all spaces in the string.

test code:
```c
#include <stdio.h>
#include <string.h>

void swoole_rtrim(char *str, int len)
{
    int i;
    for (i = len; i > 0; i--)
    {
        switch (str[i])
        {
        case ' ':
        case '\0':
        case '\n':
        case '\r':
        case '\t':
        case '\v':
            str[i] = 0;
            break;
        default:
            break;
        }
    }
}

void swoole_rtrim_new(char *str, int len)
{
    int i;
    for (i = len; i > 0; i--)
    {
        switch (str[i])
        {
        case ' ':
        case '\0':
        case '\n':
        case '\r':
        case '\t':
        case '\v':
            str[i] = 0;
            break;
        default:
            return;
        }
    }
}

void print_hex(char *str, int len)
{
    int i;
    for (i = 0; i < len; i++) {
        printf("%02X ", str[i]);
    }

    printf("\n");
}

int main(int argc, char *argv[])
{
    char str1[16] = {0};
    char str2[16] = {0};
    memcpy(str1, "a b c\r\n", 7);
    memcpy(str2, "a b c\r\n", 7);

    swoole_rtrim(str1, 15);
    swoole_rtrim_new(str2, 15);

    printf("old:\n");
    print_hex(str1, 15);
    printf("new:\n");
    print_hex(str2, 15);

    return 0;
}
```
test result:
```
old:
61 00 62 00 63 00 00 00 00 00 00 00 00 00 00
   ^^    ^^
new:
61 20 62 20 63 00 00 00 00 00 00 00 00 00 00
```

matyhtf:
You are right, I am wrong. Merged.

PR #1469, opened 2018-01-24
perfect mysql_decode_row_prepare function，now  it decode null value right.

deepziyu:
The binary protocol sends NULL values as bits inside a bitmap instead of a full byte (such as MYSQL_TYPE_NULL or byte of length is 251),
when we use COM_STMT_EXECUTE command.

@see https://dev.mysql.com/doc/internals/en/binary-protocol-value.html
@see  https://dev.mysql.com/doc/internals/en/null-bitmap.html


matyhtf:
Good job.

PR #1495, opened 2018-02-22
Fix segment fault error if passing in variables can't be serialized t…

doubaokun:
Displaying user friendly Error messages and Stack trace if passing the variables can't be serialized to Swoole\Server->task(), for example Object(Swoole\Http\Request).


matyhtf:
Should return -1

PR #1589, opened 2018-04-29
add a static method: Coroutine::getRuid

scil:
Coroutine::getRuid is based on Coroutine::getUid, but returns a uniq id for each of current requests/receive.

It will always return 1 in callbacks WorkerStart and WorkerStop.


matyhtf:
Hi @scil , Thank you for your commitment. 
This feature can be implemented using php code. Do not need extension layer support.

scil:
I think it's useful for refactoring php framework like LaravelFly

PR #1717, opened 2018-06-14
ensure C++ is used for link command

remicollet:
This may raise very strange issue, especially when building on RHEL / CentOS with gcc from devtoolset (g++ command need to be invoked at link time to produce proper binaries)

Without this change, during the make

`libtool: link: cc -shared  -fPIC -DPIC  .libs/swoole.o  ...`

With 

`libtool: link: g++  -fPIC -DPIC -shared -nostdlib /usr/lib/gcc/x86_64-redhat-linux/7/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/7/crtbeginS.o  .libs/swoole.o ...`

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -630,7 +630,7 @@ if test "$PHP_SWOOLE" != "no"; then
          LDFLAGS="$LDFLAGS -lboost_context"
     fi
 
-    PHP_NEW_EXTENSION(swoole, $swoole_source_file, $ext_shared)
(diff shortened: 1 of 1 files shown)


remicollet:
FYI, same issue was fixed the same way in PHP

https://github.com/php/php-src/commit/6af22d0d3ed224d89be1d1e9e53ff7e2e2e5439d#diff-db329e88849d2e0aea55815e5476e70a

https://bugs.php.net/bug.php?id=73956

PR #1743, opened 2018-06-24
Sync PHP examples code style.

Maikuolan:
(Just synchronizes the PHP examples code style. <strike>Nothing else changed</strike>).

*Edit: Will try to fix Codacy/PR complaints.*

### DIFF
--- a/examples/php/debug_server.php
+++ b/examples/php/debug_server.php
@@ -1,5 +1,4 @@
 <?php
-
 class DebugServer
 {
     protected $alloc_point = array();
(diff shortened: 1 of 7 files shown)


huangzhhui:
Why does not use PSR2 code style directly ? I cannot accept this.

Maikuolan:
> Why does not use PSR2 code style directly ? I cannot accept this.

:+1: 

Actually, before any changes, these files already don't follow PSR-2. My goal here, was only to make the styles already used, look consistent (e.g., sometimes there was `if (` and `if(`, or tabs + spaces in the same file, etc - these things don't look good), but also, without changing too much (so, only improving aesthetics, and not changing any functionality).

But, I agree with you. Properly following PSR-2 would be better. The files still need work though anyway, I think.

Maybe, we can close this PR (i.e., don't merge these commits), but as an alternative, run these files through a proper PSR-2 validator/fixer (e.g., [php-cs-fixer](https://cs.sensiolabs.org/))?

(Also, PSR-2 doesn't address the Codacy/PR complaints. Those would need to still be fixed too, I think).

huangzhhui:
I suggust to fix all php scripts in this repo by php-cs-fixer, it is time to get everything on track.

Maikuolan:
Good idea, I think. :+1: 

I will close this PR now.

Fixing with php-cs-fixer is (in most cases) a quick + easy task. Better for a maintainer to do that, or should I send a new PR? (I don't mind, either way).

petk:
I think PSR-2 would be very nice addition in the code examples.

Just download this:
https://github.com/FriendsOfPHP/PHP-CS-Fixer
and run php-cs-fixer fix <file-or-directory>

I think in most cases it will go through ok. :+1:

andreybolonin:
https://github.com/swoole/swoole-src/pull/1792

PR #1769, opened 2018-07-02
Replace obsolete AC_TRY_COMPILE macro with AC_COMPILE_IFELSE

petk:
Hello, Autoconf made several macros obsolete. The `AC_TRY_COMPILE` macro has been replaced by the `AC_COMPILE_IFELSE`.

Reference docs: https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Obsolete-Macros.html

This patch has been tested on Ubuntu and soon will be also done on Alpine which has POSIX shell. Autoconf made this macro obsolete a very long time ago. So, it is pretty much safe to go with the new macro instead.

Thank you for considering checking this out or merging it.


matyhtf:
Good, Thanks

PR #1771, opened 2018-07-02
support pass a file descriptor to another process

windrunner414:


matyhtf:
API design needs to be reorganized

Issue #1853, opened 2018-07-31
configure fail: "checking for /ext/sockets/php_sockets.h ... no"


diyism:
=======================configure error========================
git clone -b PHP-7.2 --depth 1 https://github.com/php/php-src.git
cd php-src/
git clone -b master --depth 1 https://github.com/swoole/swoole-src.git ext/swoole
./buildconf --force
./configure --prefix=/usr/local/php7 --disable-all --enable-cli --disable-cgi --disable-fpm --disable-phpdbg --enable-bcmath --enable-hash --enable-json --enable-mbstring --enable-mbregex --enable-mbregex-backtrack --enable-sockets --enable-pdo --with-sodium --with-password-argon2 --with-sqlite3 --with-pdo-sqlite --with-pcre-regex --with-zlib --with-openssl --enable-swoole-static --enable-openssl --with-swoole

checking /ext/sockets/php_sockets.h usability... no
checking /ext/sockets/php_sockets.h presence... no
checking for /ext/sockets/php_sockets.h... no
configure: error: cannot find /ext/sockets/php_sockets.h. Please check if sockets extension installed

=========================problem=======================

I found this:
https://github.com/swoole/swoole-src/commit/744c31cbf65bcff51740d61bdcb3b5831ec7d591#diff-788d457a20b110cc38e571dec9ddc68c 

any fix?

twose:
Ping @petk

petk:
Hm, indeed, building PECL extensions from the php-src/ext directory to statically building them into PHP directly... I think the new patch fixes this also then :) Thanks for reporting.

twose:
Fixed in #1855

PR #1855, opened 2018-08-01
Fix building extension from php-src/ext directory

petk:
Hello, this patch addresses #1853 

When building extensions directly from the php-src/ext directory, for example to build the extension statically into PHP directly, PHP header files are located in the $abs_srcdir directory.

This patch adds additional check for checking file locations of the socket extension.

Additional macro `PHP_ADD_EXTENSION_DEP` call is added for more proper initialization order. In case, sockets extension is a dependency, it must be initialized before swoole.

Probably some more additional checks should be made, but I think this should for for many cases so far. In case of issues, let me know.

Thanks.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -256,15 +256,20 @@ if test "$PHP_SWOOLE" != "no"; then
     fi
 
     if test "$PHP_SOCKETS" = "yes"; then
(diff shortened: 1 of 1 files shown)


twose:
Thanks.

PR #1878, opened 2018-08-10
compatible with solaris

CismonX:
Add some platform-specific code for compatibility with Oracle Solaris.
Now it can be built on Solaris without errors and passing most test cases.
Still stricky to build, though. You have to hack your way through the linking procedure.

Bug fixes and support for event ports are comming soon.


swoole_channel_coro.cc
```diff
@@ -36,7 +34,7 @@ typedef struct
     swLinkedList *consumer_list;
     bool closed;
     int capacity;
-    queue<zval> *data_queue;
+    std::queue<zval> *data_queue;
```
CismonX: In Solaris there's a `struct` named `queue` somewhere in its system headers, making `queue` ambiguous.
That's why we shouldn't `use namespace std;`

twose:
looks pretty good

PR #2081, opened 2018-11-02
fix co::exec

windrunner414:
If not close STDIN, some programs will be suspended
such as ffmpeg


matyhtf:
This is not compatible with PHP shell_exec, it does not close stdin

windrunner414:
if not close stdin,another pipe is needed.
Otherwise some programs will not work.
ffmpeg can't work without >/dev/null 2>&1 but it works in PHP shell_exec

PR #2112, opened 2018-11-11
Added method getState() to swoole_mysql_coro.cc

magnusvarik:
When a MySql connection object is pulled from a connection pool, there is currently no "cheap" way to check if the connection is really still there before it is injected to the application side. Being able to check SW_MYSQL_STATE_CLOSED first is very efficient and the infrastructure for this is already in place in current code.

### DIFF
--- a/swoole_mysql_coro.cc
+++ b/swoole_mysql_coro.cc
@@ -36,6 +36,7 @@ static PHP_METHOD(swoole_mysql_coro, rollback);
 static PHP_METHOD(swoole_mysql_coro, prepare);
 static PHP_METHOD(swoole_mysql_coro, setDefer);
 static PHP_METHOD(swoole_mysql_coro, getDefer);
+static PHP_METHOD(swoole_mysql_coro, getState);
 static PHP_METHOD(swoole_mysql_coro, close);
 
 static PHP_METHOD(swoole_mysql_coro_statement, __destruct);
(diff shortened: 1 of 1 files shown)


twose:
it's fault, no cheap way to check if the connection really exists, we must query or execute to communicate with database server then check the errno, if it failed, we reconnect then try again.
don't rely on connection state value.

magnusvarik:
Thank you for the feedback. Yes, the errors should be catched and handled accordingly. That's not what I had in mind and I’m sorry for not explaining my intentions. In a typical async environment with shared resources, the state can also change for other reasons than errors and closed connection is just one example.

When we have some sleeping connections in our \Swoole\Coroutine\Channel pool and intentionally restart MySql server, all the \Swoole\Coroutine\Mysql objects are returned from pool with state SW_MYSQL_STATE_CLOSED. This data is already in-memory, but not accessible to php side as we have no method for reading it. Since we just restarted the MySql service and we know it’s working, the application could just gracefully replace these closed connections as they are pulled from the pool with new ones without restarting the application or running expensive statements/queries.

Being able to check the state (not just CLOSED, but also QUERY and READ_*) in an async environment with shared connections can be very useful for many things and the getState() method was already implemented in the regular \Swoole\Mysql client module. I think it is a good idea to implement it to \Swoole\Coroutine\Mysql too for compatibility reasons.

Anyway, big thanks for the hard work on this technology, I truly believe Swoole is the best thing to happen to PHP in many years :)

flddr:
Hmm, i have done this for _redis clients_ in my own manager in about 5 lines of code. I think, it's ok to handle that myself, but i do not use _mysql_ so i do not know if it's another situation.

So, what's about like redis `$mysql->connected`?

magnusvarik:
Well, you can’t really trust this type of “isConnected” method for MySql without actually performing some kind of I/O and that’s not “cheap”. I do not recommend that. But, if the state is already closed, you can be pretty sure it is actually closed.

The power of Swoole opens up new possibilities for PHP in situations where it has traditionally not been implemented. I don’t want to sound like smart ass, but let’s say we have an application with heavy load which uses maximum amount of SQL connections during business hours. At night, the DBA have patched a vulnerability in MySql server software and performs a quick restart. All connections in our connection pool on the application side are now in state CLOSED. Since DBA does not own the application, he can’t restart it to reset them. But, it’s OK because our developer already predicted something like this would happen and implemented a simple auto-recovery for replacing closed connections as they are popped out from our Channel. This kind of check does not require any expensive I/O as it would just check the value already in Swoole memory and takes only few lines of code to implement.

twose:
`get state` as same as reading `connected` property, why not checking `connected`?
and we plan to refactor mysql_coro C code to C++ with coroutine way, after doing these, even if the server closes the connection, the client will not know, coroutine is the way to code like synchronize, if you use pdo or mysqli, you can't get the connection state anyway, unless you ask the server.

magnusvarik:
I assumed the connected property would not show other states. But if the future refactoring is going to break this anyway, then there is no point of course.
Thanks for the help! :)

flddr:
Thanks for your help :+1:

PR #2126, opened 2018-11-17
Update README.md

lesha888:
Correct url to easyswoole


twose:
Thank you for your correction, but it should be www.easyswoole.com

PR #2180, opened 2018-12-04
Modify: remove unnecessary bzero() initialization

huanghantao:
Because the following code assigns a value to the member of reactor_object.


twose:
Defensive programming, do not over optimization.

PR #2214, opened 2018-12-16
Fix unexpected swoole sever crash on high concurrency operation

jobs-git:
`SplQueue` is operated using `enqueue()` and `dequeue()` not with `push()` and `pop()` which is for `Stack`.

`SplQueue` is a `FIFO`, but use of `push/pop` turns it into `Stack` which is `LIFO`, this has caused numerous disconnections that resulted into a swoole server crash.

By fixing this, disconnection from database server does not cause server crash anymore.

Ref.

[SplQueue dequeue](http://php.net/manual/en/splqueue.dequeue.php)
[SplQueue enqueue](http://php.net/manual/en/splqueue.enqueue.php)

### DIFF
--- a/examples/coroutine/redis_pool.php
+++ b/examples/coroutine/redis_pool.php
@@ -11,13 +11,13 @@ $server->on('Request', function($request, $response) use(&$count, $pool) {
             $response->end("redis connect fail!");
             return;
         }
-        $pool->push($redis);
+        $pool->enqueue($redis);
     }
-    $redis = $pool->pop();
+    $redis = $pool->dequeue();
     $count ++;
     $ret = $redis->set('key', 'value');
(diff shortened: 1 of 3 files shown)


jobs-git:
Code in here should be fixed as well https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/swoole/swoole-server.php

flddr:
Great @jobs-git 👍

twose:
`enqueue && dequeue` is as same as `push && shift`, but this is not the reason for the disconnection,  don't mislead others.
Using pop is indeed a mistake, but any program cannot depend on the order, every time you take a connection from the connection pool, **you must check the status of the connection, even try to reconnect after the request fails**. The real network is random and uncertain, the connection will be disconnected at any time, even halfway through your request.

PR #2313, opened 2019-01-14
fix bug of updating id property of process

junwei-qu:


twose:
Thanks, but please fix file permissions, try to add a test.

PR #2323, opened 2019-01-16
fix directory traversal bug

windrunner414:


src/network/port.c
```diff
@@ -617,6 +617,22 @@ static int swPort_http_static_handler(swServer *serv, swHttpRequest *request, sw
...
+    {
+        return SW_FALSE;
+    }
+
+    if (real_path[serv->document_root_len] != '/')
+    {
+        return SW_FALSE;
+    }
+
+    if (strncmp(real_path, serv->document_root, serv->document_root_len) != 0)
```
woody712: if "document_root" is not a real path, the "document_root" is unavailable

windrunner414: fixed in #2513 , Thanks for your review

PR #2349, opened 2019-01-29
fix swoole_websocket_server exists

junwei-qu:


twose:
if it's not websocket port, it should behave as same as `Server::exist`.
This is a problematic API, it will be removed in v4.3.0, we should use `WebSocket\Server->isEstablished`.

PR #2357, opened 2019-02-01
zero copy send

jobs-git:
When sending larger content in `$response->end()` you will notice that the server performance drops significantly due to high cpu usage. This is observed even though there are almost no cpu intensive computations in the code, this indicates that direct memory access (DMA) is not being taken advantage by the server, resulting to massive degradation in performance.

By enabling zero copy, the cpu will not parse information byte by byte, thus, increasing the performance of the server.

This is still experimental as there should be an implementation for zero copy recv, error handlers and updates for other send/recv functions. Moreover, zero copy send/recv was just recently been available in Linux 4.18 and up.

Some details can be read here https://lwn.net/Articles/726917/ & https://lwn.net/Articles/752188/

### DIFF
--- a/swoole_http_server.cc
+++ b/swoole_http_server.cc
@@ -53,6 +53,8 @@ extern "C"
 #include "thirdparty/picohttpparser/picohttpparser.h"
 #endif
 
+#define enable_zero_copy 0
+
 using namespace swoole;
 
 swString *swoole_http_buffer;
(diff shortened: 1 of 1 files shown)


matyhtf:
This is invalid, server->send is not a system call, cannot use MSG_ZEROCOPY flag.

DMA This is very dangerous and may result in incomplete transmission or incorrect data being sent. The underlying memory persistence is required.

PR #2394, opened 2019-02-28
the return value of swScoket_wait_multi should be same as swSocket_wait

junwei-qu:
如果其返回值与swSocket_wait保持一致，应该会更好吧


matyhtf:
要返回可读事件个数

PR #2513, opened 2019-04-18
fix document_root can't be a relative path

windrunner414:


twose:
为什么需要支持这个?

windrunner414:
@twose 一般来说没有用
#2323

windrunner414:
cc @woody712
Are there any situations you need to use relative path?

windrunner414:
cc @woody712
Are there any situations you need to use relative path?

windrunner414:
cc @twose

PR #2589, opened 2019-05-20
Use Zend zeroing macro to clear up the sha1 context.

devnexen:
Mainly for security matter regardless of compiler optimisation level used.


matyhtf:
Cannot include php header files in code in src

PR #2612, opened 2019-06-05
fix bug: Fixed swoole_curl_error panic

gitsrc:


gitsrc:
This value of  ``$obj->errMsg`` may be null, causing the php kernel to report an error.

matyhtf:
You can modify the default value of errMsg property

gitsrc:
ok, please check #2613

PR #2772, opened 2019-08-23
Fixed allowing random HTTP methods like GETX POSTX PUTX etc

doubaokun:
At the moment, method "GETX" is treated as "GET", for example:

```php
<?php 
$server = new \Swoole\Http\Server('0.0.0.0', 9501);
$server->on('request', function ($request, $response) use ($server) {
    $response->header('Content-Type', 'text/plain');
    var_dump($request);
    $response->end("hello\n");
});
$server->start();
```

```
curl -X GET1 http://127.0.0.1:9501/ -I -v
curl -X GET2 http://127.0.0.1:9501/ -I -v
```


twose:
we need more unit tests

Issue #2743, opened 2019-08-07
File uploads $request->files do not match $_FILES


sshymko:
**Preconditions:**
- Swoole 4.x (any version)

**Steps to reproduce:**
1. Create a simple Swoole server that responds with uploaded files:
    ```php
    $server = new \Swoole\Http\Server('127.0.0.1', 8180);
    $server->on('request', function ($request, $response) {
        $response->header('Content-Type', 'text/plain');
        $response->end(var_export($request->files, true));
    });
    $server->start();
    ```
2. Create sample files with test content:
    ```bash
    echo test1 > test1.txt
    echo test22 > test2.txt
    ```
3. Upload files as **array** via `multipart/form-data` POST request:
    ```bash
    curl http://127.0.0.1:8180/ \
        -F 'form[section][file1]=@test1.txt' \
        -F 'form[section][file2]=@test2.txt'
    ```

**Expected result:**
- It is trivial to replace usage of `$_FILES` with `$request->files`
- Swoole parses uploaded files identically to `$_FILES`:
    ```php
    array (
      'form' => 
      array (
        'name' => 
        array (
          'section' => 
          array (
            'file1' => 'test1.txt',
            'file2' => 'test2.txt',
          ),
        ),
        'type' => 
        array (
          'section' => 
          array (
            'file1' => 'text/plain',
            'file2' => 'text/plain',
          ),
        ),
        'tmp_name' => 
        array (
          'section' => 
          array (
            'file1' => '/tmp/swoole.upfile.vtaZBA',
            'file2' => '/tmp/swoole.upfile.kY0ker',
          ),
        ),
        'error' => 
        array (
          'section' => 
          array (
            'file1' => 0,
            'file2' => 0,
          ),
        ),
        'size' => 
        array (
          'section' => 
          array (
            'file1' => 6,
            'file2' => 7,
          ),
        ),
      ),
    )
    ```

**Actual result:**
- It is very difficult to replace usage of `$_FILES` with `$request->files`
- Swoole parses uploaded files differently from `$_FILES`:
    ```php
    array (
      'form' => 
      array (
        'section' => 
        array (
          'file1' => 
          array (
            'name' => 'test1.txt',
            'type' => 'text/plain',
            'tmp_name' => '/tmp/swoole.upfile.vtaZBA',
            'error' => 0,
            'size' => 6,
          ),
          'file2' => 
          array (
            'name' => 'test2.txt',
            'type' => 'text/plain',
[...]

PR #2773, opened 2019-08-23
File uploads array parsing

sshymko:
- Fix file uploads data structure to accommodate array form fields #2743
- Update file uploads data structure `$request->files` to match `$_FILES`
- Test file uploads data structure of different form field nesting levels
- Add Swoole logo to test fixture images


sshymko:
All existing file upload tests and the the newly introduced test for array form fields upload pass.

twose:
nice job but I don't know whether it will cause backward incompatibility

sshymko:
@twose 
Yes, you're right the file uploads format changes for array form fields.
My arguments for accepting the changes:
- File uploads format is the same for flat form fields; it has changed only for arrays
- All existing file upload tests pass unmodified because they tested flat file uploads only
- File uploads format has never been documented, neither for flat nor array form fields
- Relying on undocumented functionality comes with backward-compatibility risks
- According to you and @matyhtf, array file uploads weren't supported (at least intentionally)

sshymko:
@twose 
Introducing a setting is a good option except it would require to maintain both formats going forward.
Since the format is not documented and only matches `$_FILES` for flat form fields, probably not many projects rely on the behavior of array form fields.

sshymko:
@twose 
Introduced setting `http_parse_files` disabled by default and covered both formats with tests.

twose:
👍great job, it looks good

twose:
How can I touch you online faster? Twitter?

PR #2783, opened 2019-08-26
Added warning message for misusing timers in server context.

doubaokun:
Several developers reported the timer function is not working because of misusing, so added the warning message.

Common mistakes:
```php
<?php
<?php declare(strict_types=1);

namespace App;

use Swoole\Http;
use Swoole\Runtime;
use Swoole\Server;
use Swoole\Timer;

Runtime::enableCoroutine();

$http_server = new Http\Server('127.0.0.1', 9000);
$http_server->on('request', function () {});

// correct example
$http_server->on('WorkerStart', function($server) {
    $server->tick(500, function () {
        echo "tick\n";
    });
});

// mistake 1
$http_server->tick(500, function () {
    echo "tick\n";
});

// mistake 2
Timer::tick(500, function () {
   echo "tick2\n";
});

$http_server->start();
```


matyhtf:
https://github.com/swoole/swoole-src/commit/7708def246a5321cbedf540f5b6a8da4211de86f

Better way

PR #2808, opened 2019-09-03
Max request settings fix

sshymko:
- Fix event worker settings having no effect
- Fix task worker settings mistakenly applying to event workers
- Cover `max_request` and `max_request_grace` settings with tests


sshymko:
@matyhtf 
Oh, you merged the PR so quickly. I was going to work on it a bit more to achieve passing tests. Let me create another PR fixing tests.

PR #2809, opened 2019-09-03
Fix max_request/max_request_grace tests

sshymko:
- Fix hang up of `max_request/max_request_grace` tests
- Remove `worker_num` repeat assignment already done in constructor

### DIFF
--- a/src/server/reactor_process.cc
+++ b/src/server/reactor_process.cc
@@ -103,7 +103,6 @@ int swReactorProcess_start(swServer *serv)
      * store to swProcessPool object
      */
     serv->gs->event_workers.ptr = serv;
-    serv->gs->event_workers.worker_num = serv->worker_num;
     serv->gs->event_workers.max_wait_time = serv->max_wait_time;
     serv->gs->event_workers.use_msgqueue = 0;
     serv->gs->event_workers.main_loop = swReactorProcess_loop;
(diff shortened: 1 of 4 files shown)


sshymko:
@matyhtf 
Let's wait for tests to pass before merging this PR.

sshymko:
@matyhtf 
I don't understand why the tests are hanging up. Could you please help to fix them?

src/server/reactor_process.cc
```diff
@@ -103,7 +103,6 @@ int swReactorProcess_start(swServer *serv)
      * store to swProcessPool object
      */
     serv->gs->event_workers.ptr = serv;
-    serv->gs->event_workers.worker_num = serv->worker_num;
```
sshymko: Assignment is already done in constructor.

twose:
PHP_INT_MAX is greater than INT_MAX, the value is overflow, and assertion failed:
```C
Assertion failed: (max > min), function swoole_system_random, file /Users/twosee/Toast/swoole-src/src/core/base.c, line 477.
```
Maybe you need a DEBUG version of PHP and Swoole

Here is what you should do (for example):
https://github.com/swoole/swoole-src/blob/b3e0f8331c1fd73602633f7fec167fdc84582bb5/swoole_async_coro.cc#L104

and I prefer to use size_t/sszie_t rather than int/uint

I don't understand why to use PHP_INT_MAX instead of other values ​​to represent infinity?

twose:
try to use uint32_t and use SW_MAX/SW_MIN to limit it

sshymko:
@twose 
Thanks for your help! Option `max_request_grace` uses the same integer types as `max_request`. Replacing `PHP_INT_MAX` with a lower value within 32-bit int range solved the issue with tests.

PR #2844, opened 2019-09-20
Performance optimization for Channel

twose:
Code
---
```php
$s = microtime(true);
Swoole\Coroutine\run(function () {
    $channel = new Swoole\Coroutine\Channel(10000);
    Swoole\Coroutine::create(function () use ($channel) {
        for ($n = 1000 * 10000; $n--;) {
            $channel->push(true);
        }
    });
    Swoole\Coroutine::create(function () use ($channel) {
        for ($n = 1000 * 10000; $n--;) {
            assert($channel->pop() === true);
        }
    });
});
$s = microtime(true) - $s;
var_dump($s);
```

Before
---
float(16.383640050888)

After
---
float(10.795506000519)

### DIFF
--- a/src/coroutine/channel.cc
+++ b/src/coroutine/channel.cc
@@ -63,27 +63,38 @@ void* Channel::pop(double timeout)
     }
     if (is_empty() || !consumer_queue.empty())
     {
-        timer_msg_t msg;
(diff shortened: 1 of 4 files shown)


twose:
BC break

PR #2891, opened 2019-10-17
Close in non current worker

matyhtf:


src/server/master.cc
```diff
@@ -1307,21 +1307,32 @@ static int swServer_tcp_close(swServer *serv, int session_id, int reset)
...
+    swDataHead ev = { 0 };
+
+    if (swServer_dispatch_mode_is_mod(serv))
+    {
+        int worker_id = swServer_worker_schedule(serv, conn->fd, nullptr);
+        if (worker_id != (int) SwooleWG.id)
+        {
+            worker = swServer_get_worker(serv, worker_id);
+            goto _notify;
+        }
```
twose: 此处retval为未初始化的值

PR #3070, opened 2020-01-17
fix: segmentation fault

huanghantao:
When the worker process execute the chunk logic,
the package will be released, but this variable is useless


huanghantao:
我好像知道package的意思了，之前应该是没有测试过chunk，所以留了这个bug

PR #3291, opened 2020-05-05
Add curl options

sy-records:


matyhtf:
No longer accept the merger request of the library, there will be a core maintainer to update uniformly before issuing the version

PR #3419, opened 2020-06-23
Replace `bzero()` calls with `memset()`

CismonX:
The POSIX function `bzero()` is deprecated in IEEE Std 1003.1-2001 and **removed** in IEEE Std 1003.1-2008. 

Although most platforms reserve it for compatibility, it is not guarenteed to exist, even in `<strings.h>`.

It is recommented that we replace that call with `memset()` instead, which is a standard ISO C function:
```C
#define sw_bzero(s, n)         memset(s, '\0', n)
```

### DIFF
--- a/core-tests/src/memory/ringbuffer.cpp
+++ b/core-tests/src/memory/ringbuffer.cpp
@@ -57,7 +57,7 @@ static void thread_write(void)
     uint32_t size, yield_count = 0, yield_total_count = 0;
     void *ptr;
     pkg send_pkg;
-    bzero(&send_pkg, sizeof(send_pkg));
+    sw_memset_zero(&send_pkg, sizeof(send_pkg));
 
     int i;
     for (i = 0; i < WRITE_N; i++)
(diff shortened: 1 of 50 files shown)


twose:
I think it is better to directly replace `bzero` to `memset` than to use macro replacement

matyhtf:
`sw_memset_zero` 

Is this name better?

matyhtf:
The second parameter of the memset function is useless

CismonX:
@matyhtf 

Yeah, `sw_memset_zero` is more straightforward.

CismonX:
@matyhtf 

The second parameter of `memset()` is the value used to fill the byte string.

Here we only want a `bzero()` alternative, so hard-code it to zero in the macro should be okay.

matyhtf:
It is all used to initialize to 0 in swoole-src, the memset() needs to input 3 parameters, resulting in repetitive code, so swoole has always used bzero instead of memset

CismonX:
@matyhtf 

Yes. That's why I introduced a new macro in this PR, rather than directly replace the `bzero()` calls with `memset()`, like @twose just said.

PR #3427, opened 2020-06-28
fix: fix build error

huanghantao:
without `sys/random.h`, `core-tests` will fail to compile.


huanghantao:
without `sys/random.h`, `core-tests` will fail to compile.

PR #3440, opened 2020-07-02
Fix the result of stream_select() is not equal to $read + $write + $error

Yurunsoft:


tests/swoole_runtime/stream_select/return_val.phpt
```diff
@@ -0,0 +1,48 @@
...
+    }
+    proc_terminate($proc);
+    if (defined('SIGKILL')) {
+        proc_terminate($proc, SIGKILL);
+    } else {
+        proc_terminate($proc);
+    }
+    proc_close($proc);
+});
+Swoole\Event::wait();
```
huanghantao, line 43: We recommend that using `Swoole\Coroutine\run` instead of `go + Swoole\Event::wait`

PR #3487, opened 2020-07-22
Support yield when sleep time < 1ms

Yurunsoft:
I recently used a package, usleep() is used in an infinite loop, and the time is <1ms, which causes the process to be blocked

3 changes:

* Support yield when sleep time < 1ms
* Fix return value of time_nanosleep() when success
* Add time_sleep_until test

### DIFF
--- a/src/reactor/epoll.cc
+++ b/src/reactor/epoll.cc
@@ -182,12 +182,8 @@ static int swReactorEpoll_wait(swReactor *reactor, struct timeval *timeo) {
     int max_event_num = reactor->max_event_num;
     struct epoll_event *events = object->events;
 
-    if (reactor->timeout_msec == 0) {
-        if (timeo == nullptr) {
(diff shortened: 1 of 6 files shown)


huanghantao:
Maybe you should add a test to see if the coroutine has yield.

src/reactor/epoll.cc
```diff
@@ -182,12 +182,8 @@ static int swReactorEpoll_wait(swReactor *reactor, struct timeval *timeo) {
     int max_event_num = reactor->max_event_num;
     struct epoll_event *events = object->events;
 
-    if (reactor->timeout_msec == 0) {
-        if (timeo == nullptr) {
-            reactor->timeout_msec = -1;
-        } else {
```
huanghantao, line 188: This change is confusing to me. Why change epoll-related code?

PR #3489, opened 2020-07-22
fix http client download filename bug

Yurunsoft:
I found the bug while developing code for YurunHttp


huanghantao:
I don't understand what bug this fixes, and there seems to be some useless code in the test code, such as this C1 class.

PR #3496, opened 2020-07-23
Added info on SSL certs

asheroto:


huanghantao:
@asheroto 

thank you! It probably should have been in the document, not the README. We already have an introduction to this, which you can find here:

https://www.swoole.co.uk/docs/modules/swoole-server/configuration

asheroto:
> @asheroto
> 
> thank you! It probably should have been in the document, not the README. We already have an introduction to this, which you can find here:
> 
> https://www.swoole.co.uk/docs/modules/swoole-server/configuration

Ahhh I did not even know about that page...

PR #3497, opened 2020-07-24
Refator timer

matyhtf:


include/swoole_timer.h
```diff
@@ -46,62 +48,87 @@ struct swTimer_node {
...
+
+    inline TimerNode *get_ex(long id, const enum swTimer_type type) {
+        swTimer_node *tnode = get(id);
+        return (tnode && tnode->type == type) ? tnode : nullptr;
+    }
+
+    inline bool remove_by_id(long id) {
+        return remove(get(id));
+    }
+
```
huanghantao, line 122: This function should be able to be defined as:

```cpp
inline bool remove(long id)
```

include/swoole_timer.h
```diff
@@ -46,62 +48,87 @@ struct swTimer_node {
...
-    return msec1 + msec2;
+
+    inline uint64_t get_round() {
+        return round;
+    }
+
+    inline TimerNode *get_ex(long id, const enum swTimer_type type) {
+        swTimer_node *tnode = get(id);
+        return (tnode && tnode->type == type) ? tnode : nullptr;
+    }
```
huanghantao, line 117: This function should be able to be defined as:

```cpp
inline TimerNode * get_timer_node(long id, const enum swTimer_type type)
```

src/core/timer.cc
```diff
@@ -47,238 +47,224 @@ static int clock_gettime(clock_id_t which_clock, struct timespec *t) {
...
-    if (timer->map) {
-        for (auto iter = timer->map->begin(); iter != timer->map->end(); iter++) {
-            auto tnode = iter->second;
-            delete tnode;
-        }
-        delete timer->map;
+    for (auto iter = map.begin(); iter != map.end(); iter++) {
+        auto tnode = iter->second;
+        delete tnode;
     }
```
huanghantao, line 109: May be we should delete map?

include/swoole_timer.h
```diff
@@ -46,62 +48,87 @@ struct swTimer_node {
...
-    struct timeval now;
-    if (!SwooleTG.timer) {
-        return SW_ERR;
+    bool init();
+    TimerNode *add(long _msec, bool persistent, void *data, const swTimerCallback &callback);
+    bool remove(TimerNode *tnode);
+    void reinit(swReactor *reactor);
+    int select();
+
+    inline TimerNode *get(long id) {
```
huanghantao, line 97: Maybe it would be better called that:

```cpp
inline TimerNode *get_timer_node(long id)
```

Because we might get something else later on.

src/core/timer.cc
matyhtf: Not a pointer, no need to delete

PR #3498, opened 2020-07-26
Refactor socket

matyhtf:


include/swoole_socket.h
```diff
@@ -161,107 +166,122 @@ struct Socket {
...
+        case 0:
+            return SW_WAIT;
+        default:
+            return SW_ERROR;
+        }
     }
 
-    return SW_OK;
-}
+    static inline int get_domain_and_type(enum swSocket_type type, int *sock_domain, int *sock_type) {
```
huanghantao, line 247: Perhaps passing references is more expressive

matyhtf: Follow the google specification, if it is a write operation to pass pointer, read-only operation to pass reference

PR #3536, opened 2020-08-03
Update README.

santalex:
benchmark/benchmark.php does not exist.


sy-records:
移到独立仓库了，路径是这里https://github.com/swoole/benchmark/blob/master/benchmark.php

santalex:
> 移到独立仓库了，路径是这里https://github.com/swoole/benchmark/blob/master/benchmark.php

readme不需要更新链接地址吗?

PR #3539, opened 2020-08-04
Fixed: fix header inject when use CRLF

huanghantao:
The specific error can be seen: https://hackmd.io/@chromium1337/S1iaq7r-w

### DIFF
--- a/swoole_http_response.cc
+++ b/swoole_http_response.cc
@@ -501,7 +501,7 @@ static ssize_t http_build_trailer(http_context *ctx, swString *response) {
             if (UNEXPECTED(!key || ZVAL_IS_NULL(zvalue))) {
                 continue;
             }
-            
+
             if (!ZVAL_IS_NULL(zvalue)) {
                 zend::String str_value(zvalue);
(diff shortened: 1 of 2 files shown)


chromium1337:
It seems that the check is not enforced when `ucwords` set to false.
Also, would you please confirm that CRLF injection in `rawCookie` method has been fixed in this commit?

chromium1337:
Thanks for merging my PR.
Please be aware that the same security check should also be implemented in `rawCookie` method.

Test script:
```php
#!/usr/bin/env php
<?php

declare(strict_types=1);

$http = new Swoole\Http\Server("0.0.0.0", 9501);
$http->on(
    "request",
    function (Swoole\Http\Request $request, Swoole\Http\Response $response) {
        if(isset($request->get['lang'])){
            $response->rawCookie('lang', $request->get['lang']);
        }
        $response->end(highlight_file(__FILE__, true));
    }
);
$http->start();
```

huanghantao:
@chromium1337 Thanks! we will fix it later.

PR #3541, opened 2020-08-05
Fix CRLF injection in `redirect` and `header`

chromium1337:
This one should correctly patch the vulnerability in `redirect` and `header` method. The previous one (#3539) is incomplete.


huanghantao:
Thanks!

Issue #1828, opened 2018-07-23
headers sent when ob start enabled


xrjxs:
## 1. the demo code below

```
$http = new swoole_http_server("0.0.0.0", 9501);
$http->on('request', function ($request, $response) {
    $headers_sent = 'false';
    if (headers_sent()) {
        $headers_sent = 'true';
    }

    ob_start();
    echo 'headers sent: ';
    $result = ob_get_contents();
    ob_end_clean();

    $response->end($result . $headers_sent);
});
$http->start();
```

## 2. the issue:
open the browser with address http://127.0.0.1:9501
browser output: 
headers sent: false

then refresh the page, the output will always is:
headers sent: true

it means headers already sent. for example session can't start:
session_start(): Cannot send session cookie - headers already sent

it seems ob_start() caused the headers_sent return true, when I remove ob_start,
headers_sent return false (that is right)

is it an issue ? what can I do for session_start with ob_start ?

## 3. php --ri swoole

swoole
swoole support => enabled
Version => 4.0.3
Author => Swoole Group[email: team@swoole.com]
coroutine => enabled
epoll => enabled
eventfd => enabled
timerfd => enabled
signalfd => enabled
spinlock => enabled
rwlock => enabled
async http/websocket client => enabled
pcre => enabled
zlib => enabled
mutex_timedlock => enabled
pthread_barrier => enabled
futex => enabled

Directive => Local Value => Master Value
swoole.aio_thread_num => 2 => 2
swoole.display_errors => On => On
swoole.use_namespace => On => On
swoole.use_shortname => On => On
swoole.fast_serialize => Off => Off
swoole.unixsock_buffer_size => 8388608 => 8388608

twose:
You can't use `headers_sent `, it's not `PHP-FPM`.
You should use `swoole_http_request` and `swoole_http_response`
```php
function swoole_http_response->header(string $key, string $value, bool $ucwords = true);
```

xrjxs:
yes I see. but I want to use the php builtin session_start(), and with ob_start(). session_start will check the headers sent.
I can found that the swoole-framework also supported to use the builtin session_start

xrjxs:
@twose I tested session_* can use in the swoole and works prefect. but not good with ob_start.

if ($_GET['seesion_cookie']) {
  session_id($_GET['session_cookie']);
  session_start();
   // also can do this below if session_id not working
   //$fileName = session_save_path() . "/sess_$sessionId";
[...]

PR #3552, opened 2020-08-07
Fix headers_sent() with output buffering in coroutines

sshymko:
- Test output buffering `ob_*()` functions
- Assert invariance of `headers_sent()` state

### DIFF
--- a/swoole_coroutine.cc
+++ b/swoole_coroutine.cc
@@ -481,14 +481,24 @@ inline void PHPCoroutine::restore_og(php_coro_task *task) {
     }
 }
 
+inline void PHPCoroutine::save_sg(php_coro_task *task) {
+    task->headers_sent = SG(headers_sent);
+}
+
+inline void PHPCoroutine::restore_sg(php_coro_task *task) {
+    SG(headers_sent) = task->headers_sent;
+}
+
 void PHPCoroutine::save_task(php_coro_task *task) {
     save_vm_stack(task);
     save_og(task);
+    save_sg(task);
 }
 
 void PHPCoroutine::restore_task(php_coro_task *task) {
     restore_vm_stack(task);
     restore_og(task);
+    restore_sg(task);
 }
 
 void PHPCoroutine::on_yield(void *arg) {
@@ -525,9 +535,7 @@ void PHPCoroutine::on_close(void *arg) {
     }
 
     if (OG(handlers).elements) {
-        if (OG(active)) {
-            php_output_end_all();
-        }
+        php_output_end_all();
         php_output_deactivate();
         php_output_activate();
     }
@@ -614,6 +622,8 @@ void PHPCoroutine::main_func(void *arg) {
         EG(error_handling) = EH_NORMAL;
         EG(exception_class) = nullptr;
         EG(exception) = nullptr;
+        
+        SG(headers_sent) = 0;
 
         task->output_ptr = nullptr;
         task->array_walk_fci = nullptr;
--- a/swoole_coroutine.h
+++ b/swoole_coroutine.h
@@ -59,6 +59,7 @@ struct php_coro_task
     zend_error_handling_t error_handling;
     zend_class_entry *exception_class;
     zend_object *exception;
+    unsigned char headers_sent;
     zend_output_globals *output_ptr;
     /* for array_walk non-reentrancy */
     php_swoole_fci *array_walk_fci;
@@ -211,6 +212,8 @@ protected:
     static inline void restore_vm_stack(php_coro_task *task);
     static inline void save_og(php_coro_task *task);
     static inline void restore_og(php_coro_task *task);
+    static inline void save_sg(php_coro_task *task);
+    static inline void restore_sg(php_coro_task *task);
     static inline void save_task(php_coro_task *task);
     static inline void restore_task(php_coro_task *task);
     static void on_yield(void *arg);
--- a/tests/swoole_http_server/headers_sent.phpt
+++ b/tests/swoole_http_server/headers_sent.phpt
@@ -1,0 +1,49 @@
+--TEST--
+swoole_http_server: headers sent (coroutine disabled)
+--SKIPIF--
+<?php require __DIR__ . '/../include/skipif.inc'; ?>
+--FILE--
+<?php
+require __DIR__ . '/../include/bootstrap.php';
+
+$pm = new ProcessManager;
+
+$pm->parentFunc = function () use ($pm) {
+    $ch = curl_init();
+    curl_setopt($ch, CURLOPT_URL, "http://127.0.0.1:{$pm->getFreePort()}/");
+    curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
+    echo curl_exec($ch);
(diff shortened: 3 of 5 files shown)


twose:
Not quite sure what the meaning of this test is?

sshymko:
@twose 
The function `headers_sent()` should return `false` as a prerequisite for `session_*()` functions. The output buffering is the mechanism that is supposed to prevent "sending the output" along with marking "headers sent".

Currently, the failing test indicates that the issue #1828 is not fully fixed. Not sure what happened in the Swoole versions 4.0.2 -> 4.5.x, but all these functions used to work fine. Maybe I'm missing something, but I clearly had all the `session_*()` functions working alongside the PHTML template engine that uses the output buffering `ob_*()` functions.

sshymko:
@twose 
I investigated more and turns out the behavior depends on the coroutine settings.
The test passes with `enable_coroutine = false` and fails otherwise.

Will update the tests to cover both scenarios. Will rename the tests to be about the `headers_sent()` function, not the output buffering.

sshymko:
@twose 
Currently, the tests reveal inconsistency of the `headers_sent()` / output buffering behavior depending on whether the coroutines are enabled or disabled. Would you like me to update the coroutine test to PASS to match the current behavior? The question comes down to whether you'd like the fix to be made here or separately. The issue seems very intricate, so I'm not going to be able to develop the fix myself and will need help from your team.

```
[PASS] swoole_http_server: headers sent (coroutine disabled) [swoole_http_server/headers_sent.phpt] 
TEST 363/969              
========DIFF========
004+ headers_sent=1
004- headers_sent=0
========DONE========
[FAIL] swoole_http_server: headers sent (coroutine enabled) [swoole_http_server/headers_sent_coroutine.phpt] 
```

swoole_coroutine.cc
```diff
@@ -525,9 +535,7 @@ void PHPCoroutine::on_close(void *arg) {
     }
 
     if (OG(handlers).elements) {
-        if (OG(active)) {
-            php_output_end_all();
```
sshymko, line 529: Function `php_output_end_all()` does call `OG(active)`.

sshymko:
@twose 
Looks like I was able to fix the `headers_sent()` incompatibility with coroutines.
Could you please review the implementation and let me know whether you'd like any changes?

twose:
I still can't understand
The functions you mentioned are all unique to FPM server mode
Coroutine and HTTP have no relation with each other
Compatible with output_buffer is reasonable as if each process has its own output target
But HTTP is only something related to your application
e.g. I can use coroutines to create many applications which have nothing to do with HTTP

For the session-related part, I did not understand at all, maybe you can provide a realistic example
(Although I think we should disable the `Session` module in swoole)

sshymko:
@twose 
Let me try to to elaborate on the use case in more details.

You're right that the function `headers_sent()` is generally irrelevant to the Swoole environment. Applications written specifically for Swoole would use `\Swoole\Http\Request` instance (that doesn't have an equivalent method, but that's okay).

The important thing about `headers_sent()` is that it affects the built-in session functions. You're right that applications written specifically for Swoole should use a different coroutine-aware mechanism for sessions, such as [swoft-cloud/swoft-session](https://github.com/swoft-cloud/swoft-session) or [itxiao6/session](https://github.com/itxiao6/session). However, not all applications are written from scratch with considerations of the Swoole environment specifics. They can rely on the existing libraries/frameworks in the PHP ecosystem not designed with the Swoole compatibility in mind.

Some projects may attempt to retrofit an existing FastCGI/PHP-FPM application for compatibility with Swoole to achieve performance improvements by reusing bootstrapped application state in memory between requests. We have adopted [magento/magento2](https://github.com/magento/magento2) to run on Swoole to achieve [~35/45% speed up](https://twitter.com/SergiiShymko/status/1272579161924624386?s=20) for frontend and Web API requests respectively. We do disable the coroutines and successfully use the built-in PHP sessions. They work fine and allow for such projects to exist.

The problem is that there's an inconsistency in the `headers_sent()` behavior in combination with the output buffering. Everything works as expected when the coroutines are disabled: capturing the output via the `ob_*()` functions maintains the `headers_sent = 0` which satisfies preconditions of the sessions. However, the same exact code works differently with coroutines: the first request has `headers_sent = 0` but all subsequent ones have `headers_sent = 1`. There's no logical reason for such difference in the `headers_sent` flag behavior. That's why I'm proposing a simple fix for coroutines.

PLEASE, PLEASE, PLEASE do not disable the session module as that would be a show stopper for the "classical" applications, such as Magento 2, from ever running on Swoole. If anything, it's better to just decline this PR and keep the existing sessions working as they are w/o coroutines. I still think the simple `headers_sent` fix would be appropriate.

Please let me know if you have further questions. Open to discuss further offline.

twose:
Do not worry, I understand what you're saying now, I will try to solve it in a reasonable way

twose:
And, can you provide a simple session example to reproduce the problem you mentioned?
I am not familiar with it, I am some busy recently, It can save me some time

sshymko:
@twose 
A basic session example came up in a very simple library (just one small class) for compatibility of the native PHP sessions with Swoole. The bug report https://github.com/upscalesoftware/swoole-session/issues/1 demonstrates how the output buffering by Twig breaks the sessions when the coroutines are enabled. However, the issue is more generic and can be reproduced with a plain output buffering w/o template engine.

Sessions are not working when coroutines are enabled:
```php
use Upscale\Swoole\Session\SessionMiddleware;

$server = new Swoole\HTTP\Server('127.0.0.1', 8080);
$server->on('request', new SessionMiddleware(function ($request, $response) {
    ob_start();
//    echo "Test";
    ob_end_clean();

    $_SESSION['data'] = $_SESSION['data'] ?? rand();
    $response->end($_SESSION['data']);
}));
$server->start();
```
```bash
curl -c cookies.txt -b cookies.txt http://127.0.0.1:8080/
215539912

curl -c cookies.txt -b cookies.txt http://127.0.0.1:8080/
PHP Warning:  session_id(): Cannot change session id when headers already sent in vendor/upscale/swoole-session/src/SessionMiddleware.php on line 62
PHP Warning:  session_start(): Cannot start session when headers already sent in vendor/upscale/swoole-session/src/SessionMiddleware.php on line 63
PHP Warning:  session_id(): Cannot change session id when headers already sent in vendor/upscale/swoole-session/src/SessionMiddleware.php on line 80
```

The same implementation works fine when coroutines are disabled:
```php
use Upscale\Swoole\Session\SessionMiddleware;

$server = new Swoole\HTTP\Server('127.0.0.1', 8080);
$server->set([
    'enable_coroutine' => false,
]);
$server->on('request', new SessionMiddleware(function ($request, $response) {
    ob_start();
//    echo "Test";
    ob_end_clean();

    $_SESSION['data'] = $_SESSION['data'] ?? rand();
    $response->end($_SESSION['data']);
}));
$server->start();
```
```bash
curl -c cookies.txt -b cookies.txt http://127.0.0.1:8080/
2031979775

curl -c cookies.txt -b cookies.txt http://127.0.0.1:8080/
2031979775
```

PR #3720, opened 2020-10-08
Support saving server stats to file every 1 second

Yurunsoft:


doubaokun (commented):
This feature should be done with PHP in application layer.

matyhtf:
goto https://github.com/swoole/swoole-src/pull/3725

PR #3762, opened 2020-10-22
Add ext decoder

matyhtf:


huanghantao:
Perhaps, unit testing can be combined with pressure testing?

matyhtf:
> Perhaps, unit testing can be combined with pressure testing?

I think it’s better to test separately

PR #3837, opened 2020-11-13
Fix coverity scan issues

huanghantao:


ext-src/swoole_client.cc
```diff
@@ -524,10 +540,14 @@ void php_swoole_client_check_setting(Client *cli, zval *zset) {
...
                 } else {
                     php_swoole_fatal_error(E_WARNING, "http_proxy_password should not be null");
+                    // because we do not set last errcode, return true
+                    return true;
                 }
             }
         } else {
             php_swoole_fatal_error(E_WARNING, "http_proxy_port should not be null");
+            // because we do not set last errcode, return true
+            return false;
```
twose, line 550: typo here?

twose (changes requested):

ext-src/swoole_client.cc
huanghantao: > typo here?

where?

Issue #3884, opened 2020-11-26
channel空时pop操作不会阻塞


uu-myheart:
Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a simple script for reproducing the error.
![image](https://user-images.githubusercontent.com/18180275/100349825-fb1e7d00-3023-11eb-91e2-3df4729501f7.png)

2. What did you expect to see?
pop操作阻塞，而不是一直循环

3. What did you see instead?
一直在执行循环，无限打印16行的 var_dump($data);

4. What version of Swoole are you using (show your `php --ri swoole`)?
![image](https://user-images.githubusercontent.com/18180275/100350055-48025380-3024-11eb-9a2e-fbd34d9cc8bd.png)

5. What is your machine environment used (show your `uname -a` & `php -v` & `gcc -v`) ?
php官方docker镜像 php:7.2-alpine 通过 pecl install swoole 安装的 4.5.7版本
![image](https://user-images.githubusercontent.com/18180275/100350118-636d5e80-3024-11eb-84b4-ac2fff077253.png)

matyhtf:
https://github.com/swoole/swoole-src/pull/3911/files

matyhtf:
这是因为 eventloop 已经退出，在 shutdown function 里 channel 析构函数中执行了 close 操作，这会 resume 已经泄露的协程。而 channel->pop 在已关闭时会立即返回 false，造成了死循环。

Issue #3703, opened 2020-09-25
Lose current coroutine when after channel->pop()


Yurunsoft:
Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a simple script for reproducing the error.

```php
<?php
Co\run(function(){
    $channel = new Swoole\Coroutine\Channel(1);
    var_dump('cid:' . Swoole\Coroutine::getCid());
    var_dump($channel->pop());
    var_dump('cid:' . Swoole\Coroutine::getCid());
});
```

2. What did you expect to see?

```
string(5) "cid:1"
bool(false)
string(6) "cid:1"
```

3. What did you see instead?

```
string(5) "cid:1"
bool(false)
string(6) "cid:-1"
```

4. What version of Swoole are you using (show your `php --ri swoole`)?

4.5.4

5. What is your machine environment used (show your `uname -a` & `php -v` & `gcc -v`) ?

PHP 7.4.10

matyhtf:
The deadlock coroutine is awakened in the destructor, and the next version will directly throw an error and exit the process.

matyhtf:
https://github.com/swoole/swoole-src/pull/3911

PR #3911, opened 2020-12-04
Co deadlock detect

matyhtf:


matyhtf:
fix https://github.com/swoole/swoole-src/issues/3884
fix https://github.com/swoole/swoole-src/issues/3703

PR #4000, opened 2021-01-15
Optimize enable openssl

huanghantao:


huanghantao:
PHP7.2/7.3 do not define `PKG_CHECK_MODULES ` in `build/pkg.m4`

PR #4080, opened 2021-03-05
NetBSD build fix proposal.

devnexen:
strengthen gethostbyname2_r detection which symbol is indeed present in the libc but not exposed.
there is little different for the kqueue API, the private kqueue data is an integral instead.


matyhtf:
Thanks for your contribution. Great job!

PR #4116, opened 2021-03-24
enable_reuse_port directive enabled for other systems.

devnexen:


matyhtf:
Under the freebsd platform, the REUSE_PORT option has a different effect on linux. 

This is a performance optimization option for linux only

PR #4127, opened 2021-04-01
make context darwin arm64: removes workaround and uses the label

devnexen:
directly. LLVM support those since quite some years now.


devnexen:
note I upstreamed this change just today.

twose (approved):

PR #4199, opened 2021-05-11
get rid of inet_ntoa and inet_aton calls

remicollet:
These functions are deprecated and may be forbidden in some distro.

Already erraticated in php-src

* https://github.com/php/php-src/commit/f9547f2b47633d30087703df04f7d8e20f89ec80
* https://github.com/php/php-src/commit/e5b6f43ec7813392d83ea586b7902e0396a1f792
* https://github.com/php/php-src/commit/99d67d121acd4c324738509679d23acaf759d065

third-party file seems a copy paste from ph-src, perhaps need a full refresh

### DIFF
--- a/ext-src/swoole_client.cc
+++ b/ext-src/swoole_client.cc
@@ -1210,7 +1210,12 @@ static PHP_METHOD(swoole_client, getsockname) {
         }
     } else {
         add_assoc_long(return_value, "port", ntohs(cli->socket->info.addr.inet_v4.sin_port));
-        add_assoc_string(return_value, "host", inet_ntoa(cli->socket->info.addr.inet_v4.sin_addr));
+        char tmp[INET_ADDRSTRLEN];
+        if (inet_ntop(AF_INET, &cli->socket->info.addr.inet_v4.sin_addr, tmp, sizeof(tmp))) {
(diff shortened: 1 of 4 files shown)


ext-src/swoole_client.cc
```diff
@@ -1210,7 +1210,12 @@ static PHP_METHOD(swoole_client, getsockname) {
         }
     } else {
         add_assoc_long(return_value, "port", ntohs(cli->socket->info.addr.inet_v4.sin_port));
-        add_assoc_string(return_value, "host", inet_ntoa(cli->socket->info.addr.inet_v4.sin_addr));
+        char tmp[INET_ADDRSTRLEN];
+        if (inet_ntop(AF_INET, &cli->socket->info.addr.inet_v4.sin_addr, tmp, sizeof(tmp))) {
+            add_assoc_string(return_value, "host", tmp);
+        } else {
+            php_swoole_fatal_error(E_WARNING, "inet_ntop() failed");
+        }
```
remicollet, line 1218: same / consistent behavior than the below one for IPv6

remicollet:
Notice: in php-src new call usage are conditional (#ifdef HAVE_NET...)  but are already used in swoole without such condition
(btw, probably worth to be clean in php-src)

PR #4300, opened 2021-07-05
Align atomic pointers addresses

### LINKED ISSUE #4263: swoole atomic long 在M1芯片下无反应

dixyes:
gcc __sync_xxx implements at some architecture needs an aligned address to be passed in, but an allocated address may not be aligned, this patch will align it at atomic implement

this should fix #4263, but not tested at m1 chips (only at arm64 linux).


matyhtf:
https://github.com/swoole/swoole-src/commit/e6fae2ed15e26b049cc2ec25ff18b8c4d6d11117

PR #4330, opened 2021-07-20
Support a new concurrency mode

doubaokun:
* Allow applications to control max concurrent requests on each worker (Linux) process
* When the `max_concurrency = 1`, global variables and static variables are considered safe for the current request
* Allow `Laravel Octane` to use coroutines at the request level to reduce the latency

### DIFF
--- a/ext-src/php_swoole.cc
+++ b/ext-src/php_swoole.cc
@@ -384,6 +384,9 @@ void php_swoole_set_global_option(HashTable *vht) {
     if (php_swoole_array_get_value(vht, "socket_timeout", ztmp)) {
         Socket::default_read_timeout = Socket::default_write_timeout = timeout_format(ztmp);
     }
+    if (php_swoole_array_get_value(vht, "max_concurrency", ztmp)) {
+        SwooleG.max_concurrency = (uint32_t) SW_MAX(0, zval_get_long(ztmp));
+    }
 }
 
 void php_swoole_register_rshutdown_callback(swoole::Callback cb, void *private_data) {
--- a/ext-src/swoole_coroutine.cc
+++ b/ext-src/swoole_coroutine.cc
@@ -631,6 +631,10 @@ void PHPCoroutine::on_close(void *arg) {
         efree(task->array_walk_fci);
     }
 #endif
+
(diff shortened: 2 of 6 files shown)


binaryfire:
Hi all. The Laravel team is open to implementing this but they need help with the PR.  See this conversation: https://github.com/laravel/octane/issues/765

Do you know anyone who could help?

deminy:
The configuration option `max_concurrency` was added for Laravel Octane. However, it's not included in the source code of Octane directly. On the other hand, Octane has coroutine [disabled](https://github.com/laravel/octane/blob/v2.1.2/src/Commands/StartSwooleCommand.php#L122) by default, probably to avoid any issue with third-party packages and existing projects.

In the pursuit of enabling coroutines within Octane, a practical starting point is to set `max_concurrency` to 1. This approach aligns with the recommendations from Mohamed Said's [Asynchronous Laravel](https://t.co/BHVBZDhhcT) talk at Laracon Online Summer 2021. Setting `max_concurrency` to 1 allows the utilization of coroutines within a single HTTP request. However, it's crucial to understand that in this setup, each process is limited to handling one HTTP request at a time. As a result, **in many cases** a connection pool behaves more akin to a persistent connection rather than a true pool, with only one connection being utilized at any given time.

Fully leveraging Swoole's capabilities within Laravel Octane, particularly for coroutine support and handling concurrent HTTP requests, is a complex undertaking. It would likely necessitate significant refactoring. The complexity of these changes also raises concerns about their integration and acceptance within the Octane framework.

@binaryfire thanks for reaching us, and hope this helps. Please feel free to let me know if there are any questions. Thanks

binaryfire:
Thanks for your insights @deminy. The amount of refactoring needed across the whole framework to support concurrent HTTP requests would indeed be impractical. So as you said, the most practical approach would be to set `max_concurrency` to 1 and just add support for using coroutines within a single request. 

The question is, do you think that change would result in a noticeable performance improvement for high traffic apps?

deminy:
For many apps, probably not that noticeable.

High-performance apps depend on multiple aspects, not just the underline framework.

Issue #4327, opened 2021-07-20
mkdir create failed on coroutine


weskiller:
I need create many different directory, so use coroutine improve efficiency

```php
<?php

use Swoole\Coroutine;
use Swoole\Coroutine\Barrier;

use function Swoole\Coroutine\run;

Swoole\Runtime::enableCoroutine($flags = SWOOLE_HOOK_ALL);

run(function () {
    $barrier = Barrier::make();
    for ($index = 0; $index < 100; $index++) {
        Coroutine::create(static function () use ($index, $barrier) {
            if (!mkdir($concurrentDirectory = __DIR__.'/test/'.$index, 0755, true) && !is_dir($concurrentDirectory)) {
                throw new RuntimeException(sprintf('Directory "%s" was not created', $concurrentDirectory));
            }
        });
    }
    Barrier::wait($barrier);
});

```

but i get this error
```bash
#php8.0 debug.php
PHP Fatal error:  Uncaught RuntimeException: Directory "/home/weskiller/Swoole/Debug/test/1" was not created in /home/weskiller/Swoole/Debug/debug.php:15
Stack trace:
#0 {main}
  thrown in /home/weskiller/Swoole/Debug/debug.php on line 15
PHP Fatal error:  Uncaught RuntimeException: Directory "/home/weskiller/Swoole/Debug/test/3" was not created in /home/weskiller/Swoole/Debug/debug.php:15
Stack trace:
#0 {main}
  thrown in /home/weskiller/Swoole/Debug/debug.php on line 15
PHP Fatal error:  Uncaught RuntimeException: Directory "/home/weskiller/Swoole/Debug/test/2" was not created in /home/weskiller/Swoole/Debug/debug.php:15
Stack trace:
#0 {main}
  thrown in /home/weskiller/Swoole/Debug/debug.php on line 15
```

show  linux kernel and system vesion , swoole version
```bash
#uname -a && lsb_release -a && php8.0 --version && php8.0 --ri swoole
Linux dev119 5.4.0-77-generic #86-Ubuntu SMP Thu Jun 17 02:35:03 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.2 LTS
Release:	20.04
Codename:	focal
PHP 8.0.8 (cli) (built: Jul  1 2021 15:26:46) ( NTS )
Copyright (c) The PHP Group
[...]

PR #4337, opened 2021-07-24
Fix bug #4327 mkdir create failed on coroutines

NathanFreeman:
Fix bug #4327.
One of these coroutines can create top level directory successfully, but other coroutines will fail because the top directory has been created.

So I add this code to solve it. (plain_wrapper.c 1086)

```c
if (p == buf) {
    ret = mkdir(dir, mode);
} else if (!(ret = mkdir(buf, mode)) || EEXIST == errno) {
    if (!p) {
        p = buf;
    }
    /* create any needed directories if the creation of the 1st directory worked */
    while (++p != e) {
        if (*p == '\0') {
            *p = DEFAULT_SLASH;
            if ((*(p + 1) != '\0') && (ret = mkdir(buf, (mode_t) mode)) < 0) {
                // parent directory already exists and try to create child directories.
                if (EEXIST == errno && (int) strlen(buf) < dir_len) {
                    continue;
                }

                if (options & REPORT_ERRORS) {
                    php_error_docref(NULL, E_WARNING, "%s", strerror(errno));
                }
                break;
            }
        }
    }
}
```

Thanks!


matyhtf:
@NathanFreeman Good job

mifei077:
@NathanFreeman  您好，我是识沃科技的工作人员，为了感谢您的反馈，请加微信13524075254联系领取我司为您准备的定制小礼物一份，谢谢！

PR #4341, opened 2021-07-28
Support query /etc/hosts

zmyWL:


src/network/dns.cc
```diff
@@ -103,6 +106,162 @@ static int domain_encode(const char *src, int n, char *dest);
 static void domain_decode(char *str);
 static std::string parse_ip_address(void *vaddr, int type);
 
+static int read_line(FILE *fp, char **buf, size_t *bufsize) {
+    char *newbuf;
+    size_t offset = 0;
+    size_t len;
+
+    if (*buf == nullptr) {
+        *buf = (char *) malloc(128);
```
Appla, line 115: although there maybe space or comments, domain name and ip has known max size. is it possible to avoid malloc?

PR #4375, opened 2021-08-15
boost context suppoprt for riscv64

dixyes:
(Not so well) tested on hifive unmatched

- Added new boost context code from boost context and "backported" it

在hifive unmatched上（不是很完善地）测试过

- 加入了boost context的新代码并且稍微backport了下来支持swoole使用的api


dixyes:
@devnexen Could you please review this? I knew few about riscv. The fu740 cpu is slow, I almost cannot run full tests with mysql or redis (coro only part is OK).

devnexen:
> @devnexen Could you please review this? I knew few about riscv. The fu740 cpu is slow, I almost cannot run full tests with mysql or redis (coro only part is OK).

Nothing much to say since it s mostly context from boost :-) and very unfamiliar with Risc :-) it s surely good to go, boost usually does good job ;-)

PR #4544, opened 2021-12-19
Fix Z_STRLEN_P() return value is size_t

Yurunsoft:


twose:
`size_t`的format是`%zu`

PR #4551, opened 2021-12-22
Set thread name

Yurunsoft:
More convenient gdb debugging


Yurunsoft:
reference: <https://github.com/google/llvm-propeller/blob/45c226984fe8377ebfb2ad7713c680d652ba678d/lldb/test/API/tools/lldb-server/thread-name/main.cpp#L4>

matyhtf:
暂时不引入，等待思考清楚线程命名规范

PR #4559, opened 2021-12-25
Making php tests compatible with PHP8.1.

NathanFreeman:


NathanFreeman:
Unfortunately, the numbers.phpt will run failed because of  [(php bug)](https://github.com/php/php-src/pull/7837).

因为php mysqlnd的bug，单元测试numbers.phpt依旧会失败。
https://github.com/php/php-src/pull/7837

PR #4566, opened 2021-12-30
Exclude alpine 3.15 with php 8.1 for building alpine.

limingxinleo:


limingxinleo:
3.15 8.1 先排除掉吧

镜像还没制作好

https://github.com/codecasts/php-alpine/pull/154

PR #4569, opened 2022-01-03
Compiling swoole curl in different curl

NathanFreeman:


NathanFreeman:
some option are not available since 7.56.0.

PR #4572, opened 2022-01-08
Improving code coverage rate

NathanFreeman:


NathanFreeman:
Maybe I need to revert log.cc because It seem that the latest unit testing is not useful

Issue #4663, opened 2022-03-25
(Swoole Websocket) Impossible to set cookie, using default handshake method


dmitryuk:
Since [29056](https://github.com/swoole/swoole-src/pull/2906/files) Swoole supports frame compress after handshake.
I have tried it, but faced with the problem, that I cann't send Cookie on connect (session id).
Swoole onOpen event don't receives `Response` object and I see code context also has been ended. 
It's not easy to implement RFC-7692 on php

Please make the feature, for example, `on('beforehandshakeresponse')` event that will be called before https://github.com/swoole/swoole-src/blob/master/ext-src/swoole_websocket_server.cc#L384 with request and response objects, there I will can add response headers.

Please watch my PR https://github.com/swoole/swoole-src/pull/4664

NathanFreeman:
Hi!
Could you please provide the codes for reproducing the error?

NathanFreeman:
Do you mean that you can not send cookie to client server?
You can use onHandshake callback to send cookie to client server if your answer is yes.

dmitryuk:
I can't use onHandshake and websocket_compression together. 

- I can implement onHandshake (without frame compression) and set cookie
- Don't implement onHandshake and finally can't set cookie on connect

NathanFreeman:
Here is my code and work successfully.
```php
<?php
// websocker server
$server = new Swoole\Websocket\Server('127.0.0.1', 8080);
$server->set(['websocket_compression' => true]);
$server->on('handshake', function (\Swoole\Http\Request $request, \Swoole\Http\Response $response) {
    $secWebSocketKey = $request->header['sec-websocket-key'];
    $patten = '#^[+/0-9A-Za-z]{21}[AQgw]==$#';
    if (0 === preg_match($patten, $secWebSocketKey) || 16 !== strlen(base64_decode($secWebSocketKey))) {
        $response->end();
        return false;
    }
    echo $request->header['sec-websocket-key'];
    $key = base64_encode(
        sha1(
            $request->header['sec-websocket-key'] . '258EAFA5-E914-47DA-95CA-C5AB0DC85B11',
            true
        )
    );

    $headers = [
        'Upgrade' => 'websocket',
        'Connection' => 'Upgrade',
        'Sec-WebSocket-Accept' => $key,
        'Sec-WebSocket-Version' => '13',
    ];
    if (isset($request->header['sec-websocket-protocol'])) {
[...]

PR #4664, opened 2022-03-25
on('beforehandshakeresponse') event

dmitryuk:
Fixes https://github.com/swoole/swoole-src/issues/4663


NathanFreeman:
Hi @dmitryuk 
Would be also nice if you could add a new test.
Thanks.

dmitryuk:
I have added the test, but didn't find a way how to check it locally. Please accept workflow

dmitryuk:
I have checked the code in my project and it works as intended.

dmitryuk:
@NathanFreeman What we'll do next? As I see test skipped since build timeout.

NathanFreeman:
@dmitryuk 
Hi!
Sometime the build will timeout.
And next waiting for my team members to review it.
It's going to take some time.

dmitryuk:
@matyhtf please review pr 🙏

dmitryuk:
@matyhtf please approve the PR

Appla:
I don't this this is a good approach.
Is possible to check user defined onhandshake' response to determine websocket extension options which handled by swoole internally?

matyhtf:
I think both ways should be supported.

In some cases just to set a cookie or other http header, and do not want to re-implement the handshake, then we can use the beforeHandShakeResponse callback

dmitryuk:
Hello. @matyhtf still waiting for release of the feature. Can you make a new tag with PR?

PR #4807, opened 2022-08-13
Fix Bug #4794

NathanFreeman:
Adding SW_TIMER_CORRECT_TIMEOUT marcos to correct the timeout parameter if timeout is less than 0.001.


NathanFreeman:
Some tests are so extreme and I remove them.

PR #4849, opened 2022-09-28
There are some small changes to submit together

Quarkay:
- Make test sql for postgresql more robust
- Change a few code's style in swoole::coroutine::Context class (since -std=c++11)
  Advantages of pass-by-value and std::move over pass-by-reference: https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference
- Change param default value, add branch optimization information
    For all 'time<...>(true)' use cases until now, the param value is true.


src/coroutine/context.cc
```diff
@@ -31,8 +31,8 @@
 namespace swoole {
 namespace coroutine {
 
-Context::Context(size_t stack_size, const CoroutineFunc &fn, void *private_data)
-    : fn_(fn), stack_size_(stack_size), private_data_(private_data) {
+Context::Context(size_t stack_size, CoroutineFunc fn, void *private_data)
+    : fn_(std::move(fn)), stack_size_(stack_size), private_data_(private_data) {
```
matyhtf, line 35: No performance optimization, here will be copied at least once.

Quarkay: Beside the discussion link on StackOverflow above I found another document: [efficient-argument-passing-cxx11-part2](https://www.codesynthesis.com/~boris/blog/2012/06/26/efficient-argument-passing-cxx11-part2/).

> The const reference approach is efficient if we don’t make a copy of the passed argument. However, if we do, and the function is called with an rvalue, then we miss the opportunity of moving this argument instead of making a copy. So, in summary, pass by const reference is optimal if no copies are made. Otherwise, it misses out on rvalue arguments.
<br>The by-value approach is efficient if we do make a copy of the passed argument. However, if we don’t, and the function is called with an lvalue, then we make an unnecessary copy. So, in summary, pass by value is optimal if we know for sure we are going to copy the argument. Otherwise, it adds a copy overhead in case of an lvalue argument.

The parameter 'fn' is only copied to property 'fn_' in the Context::Context(...) constructor function. It's never used as a left-value in other place. And the std::function (template class) has move-constructor.

So there is an optimization case. A right-value passed as argument 'fn' then it will be moved to fn variable in Context::Context(...) function. Then moved to property fn_ again. No copy action happened. It is better than const-reference because it will make one copy action happened in this case.

matyhtf: ok, I understand, merged

PR #4877, opened 2022-10-16
Add BufferChunk's constructor and destructor

Quarkay:
- Let the two functions do what they are supposed to do, rather than mixed in Buffer Class.

- Optimize some code's style.

### DIFF
--- a/src/memory/buffer.cc
+++ b/src/memory/buffer.cc
@@ -19,34 +19,34 @@
 
 namespace swoole {
 
+BufferChunk::BufferChunk(Type type, uint32_t size) : type(type), size(size) {
+    if (type == TYPE_DATA && size > 0) {
+        value.ptr = new char[size];
+    }
+}
+
+BufferChunk::~BufferChunk() {
+    if (type == TYPE_DATA) {
+        delete[] value.ptr;
+    }
+    if (destroy) {
+        destroy(this);
+    }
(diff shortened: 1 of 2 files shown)


src/memory/buffer.cc
```diff
@@ -19,34 +19,34 @@
 
 namespace swoole {
 
+BufferChunk::BufferChunk(Type type, uint32_t size) : type(type), size(size) {
+    if (type == TYPE_DATA && size > 0) {
```
matyhtf, line 23: I think the if condition of the type should be removed, and if the size is greater than 0, the memory will be allocated

Quarkay: Yes, there is an intrinsic relationship between these two parameters according to the original design.

If the "type" conditional judgment was removed, all use cases which `type` is not `TYPE_DATA` must insure `size` equal to zero since `value` property is union type.

```cpp
BufferChunk *chunk = out_buffer->alloc(BufferChunk::TYPE_SENDFILE, 0);
chunk->value.object = task.release();
chunk->destroy = Socket_sendfile_destructor;
```

If the "type" conditional judgment was preserved, `size` would be ignored in these cases.

So I made conservative changes.

matyhtf:
@Quarkay 请加一下我的微信 19921030512

PR #4880, opened 2022-10-18
add missing HTTP2 error code, and use hexadecimal integers instead

deminy:
Two changes included in this PR:

1. Add missing HTTP2 error code `HTTP_1_1_REQUIRED ` from [RFC 7540](https://www.rfc-editor.org/rfc/rfc7540#section-7) (Hypertext Transfer Protocol Version 2 (HTTP/2)). The error code is in library `nghttp2` already.
2. Use hexadecimal integers instead. This is to use the same integer format as in RFC 7540.


deminy:
HTTP2 error codes were added by @twose to support the HTTP2 GOAWAY frame ([commit](https://github.com/swoole/swoole-src/commit/014ec2701f12c089be452dbeaced5a92b73c327a)). Not sure why he didn't have the last error code (`HTTP_1_1_REQUIRED`) included.

Requesting @twose to review this PR.

matyhtf:
Just added error codes, which can be safely merged

PR #4881, opened 2022-10-19
Fix a table crash when using getMemorySize() before add fields

Yurunsoft:


deminy:
@matyhtf The changes were merged to the 4.8.x branch. Cherry-pick to the master branch as well?

PR #4913, opened 2022-11-24
fix property source_worker_id in class PipeMessage

deminy:
Property [\Swoole\Server\PipeMessage::$source_worker_id](https://github.com/swoole/swoole-src/blob/v5.0.1/ext-src/swoole_server.cc#L490) is updated incorrectly. This PR is to fix it.


matyhtf:
The `source_worker_id` is correct, but this will generate a break change, needs to be changed with version 5.1.

matyhtf:
Adding new property will not cause backward incompatibility and can be merged into the current version

PR #5001, opened 2023-03-08
 Fix octane bug #651

matyhtf:


NargesHajiani:
@matyhtf laravel/octane#651 still exists after this PR (`Swoole v5.0.3`).

PR #5050, opened 2023-05-10
Support pdo-pgsql hook

matyhtf:


brightlycolor:
@matyhtf Hi! Is this about making native pgsql driver to by async? Is there instructions or any reference how we can use it? I use docker container

```
FROM phpswoole/swoole:5.0.3-php8.2
COPY . /app
WORKDIR /app
EXPOSE 80
ENTRYPOINT ["/app/entrypoint.sh"]
```

but it says there is no pgsql driver, try to add

```
RUN apt-get update && \
    apt-get install -y libpq-dev && \
    docker-php-ext-install pdo pdo_pgsql pgsql
```

but seems it works in sync way, maybe I have missed something and i need to enable it in swoole directly?

Appreciate any help, thank you!

Issue #5214, opened 2023-12-11
Cookie is truncated leading to a malformed/invalid cookie


stnguyen90:
Please answer these questions before submitting your issue.

1. What did you do? If possible, provide a simple script for reproducing the error.

I set a cookie as follows:

```php
    $http->on('request', function (Swoole\Http\Request $request, Swoole\Http\Response $response) use ($pm) {
        $response->cookie('test', '123456789', time() + 3600, '/', 'example.com', true, true, 'None');
        $response->end();
    });
```

2. What did you expect to see?

I expected to see a cookie set like

```
test=123456789; expires=Mon, 11-Dec-2023 21:48:04 GMT; Max-Age=3600; path=/; domain=example.com; secure; httponly; samesite=None
```

3. What did you see instead?

Instead, I saw the end was cut off like:

```
test=123456789; expires=Mon, 11-Dec-2023 21:48:04 GMT; Max-Age=3600; path=/; domain=example.com; secure; httponly; samesite=
```

4. What version of Swoole are you using (show your `php --ri swoole`)?

```shell
php --ri swoole

swoole

Swoole => enabled
Author => Swoole Team <team@swoole.com>
Version => 5.1.1
Built => Dec 11 2023 21:19:51
coroutine => enabled with boost asm context
epoll => enabled
eventfd => enabled
signalfd => enabled
cpu_affinity => enabled
spinlock => enabled
rwlock => enabled
sockets => enabled
openssl => OpenSSL 1.1.1n  15 Mar 2022
dtls => enabled
http2 => enabled
json => enabled
curl-native => enabled
pcre => enabled
c-ares => 1.17.1
zlib => 1.2.11
brotli => E16777225/D16777225
mutex_timedlock => enabled
pthread_barrier => enabled
futex => enabled
mysqlnd => enabled
async_redis => enabled
coroutine_pgsql => enabled
coroutine_odbc => enabled
coroutine_oracle => enabled
coroutine_sqlite => enabled

Directive => Local Value => Master Value
swoole.enable_coroutine => On => On
swoole.enable_library => On => On
swoole.enable_fiber_mock => Off => Off
swoole.enable_preemptive_scheduler => Off => Off
swoole.display_errors => On => On
swoole.use_shortname => On => On
swoole.unixsock_buffer_size => 8388608 => 8388608
root@d0b8cdbbff6f:/swoole-src/scripts# php --ri swoole

swoole

Swoole => enabled
[...]

PR #5215, opened 2023-12-11
Fix truncated cookie

stnguyen90:
Previously, Max-Age was added to the cookie, but cookie_size wasn't updated to account for the additional string size. This lead to the data being truncated in cases where everything was set.

This change updates the cookie_size based on the length of max_age and then increases the size of cookie to the increased cookie_size.

Fixes https://github.com/swoole/swoole-src/issues/5214

### DIFF
--- a/ext-src/swoole_http_response.cc
+++ b/ext-src/swoole_http_response.cc
@@ -987,9 +987,6 @@ static void php_swoole_http_response_cookie(INTERNAL_FUNCTION_PARAMETERS, const 
         RETURN_FALSE;
     }
 
-    size_t cookie_size = name_len /* + value_len */ + path_len + domain_len + 100;
-    char *cookie = nullptr, *date = nullptr;
-
     if (name_len > 0 && strpbrk(name, "=,; \t\r\n\013\014") != nullptr) {
         php_swoole_error(E_WARNING, "Cookie names can't contain any of the following '=,; \\t\\r\\n\\013\\014'");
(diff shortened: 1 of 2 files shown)


ext-src/swoole_http_response.cc
```diff
@@ -1042,6 +1042,11 @@ static void php_swoole_http_response_cookie(INTERNAL_FUNCTION_PARAMETERS, const
             zval max_age;
             ZVAL_DOUBLE(&max_age, diff);
             convert_to_string(&max_age);
+            cookie_size += sizeof(max_age);
+            char *new_cookie = (char *) emalloc(cookie_size);
+            strlcpy(new_cookie, cookie, cookie_size);
+            efree(cookie);
+            cookie = new_cookie;
```
NathanFreeman, line 1049:
```c++
size_t cookie_size = name_len /* + value_len */ + path_len + domain_len + 200 + samesite_len + priority_len;
```
Attempting to modify the allocated memory size here.

stnguyen90: We don't know the exact length of `max_age` until `expires` is processed, though. 🧐

stnguyen90: @NathanFreeman, I've updated the code to try and calculate the cookie_size. How is it now?

Yurunsoft: It would be better to use `sizeof("") - 1` to represent the length

PR #5239, opened 2024-01-24
Fix out of sources tree build

remicollet:
Broken by eb9a16ea613429f473b833b3ea302aefa50656d3 and 03b263900c

Notice: **pecl** build use out of sources tree build

So tagged version 5.1.2 will be unusable there


remicollet:
Notice: It will be nice to use some out of sources tree build in CI

```
phpize
mkdir cibuild
cd bibuild
../configure
make

```

Issue #5313, opened 2024-05-05
Alpine + Swoole + PgSQL on the host = `could not send SSL negotiation packet: Resource temporarily unavailable`


flexchar:
Since the issue is so mysterious, I was able to and made a reproducible. Please try it yourself.

https://github.com/flexchar/pdo-pgsql-issue

-----
Please answer these questions before submitting your issue.

1. What did you do? If possible, provide a simple script for reproducing the error.
I attempt to connect to the pgsql running on my host mac machine from inside docker container running Laravel. It would fail with `SQLSTATE[08006] [7] could not send SSL negotiation packet: Resource temporarily unavailable` error. 

It does work if I would connect using `psql` cli terminal. It also works if I used `pg_connect` instead `php-pdo`. 

I cannot test with php83 since the latest swoole pecl upstream has been broken for months now. I am not sure what's going on.

2. What did you expect to see?
It should connect.

3. What did you see instead?
`SQLSTATE[08006] [7] could not send SSL negotiation packet: Resource temporarily unavailable`

4. What version of Swoole are you using (show your `php --ri swoole`)?

```bash
Swoole => enabled
Author => Swoole Team <team@swoole.com>
Version => 5.1.1
Built => Nov 28 2023 19:24:57
coroutine => enabled with boost asm context
epoll => enabled
eventfd => enabled
signalfd => enabled
spinlock => enabled
rwlock => enabled
sockets => enabled
openssl => OpenSSL 3.1.4 24 Oct 2023
dtls => enabled
http2 => enabled
json => enabled
curl-native => enabled
c-ares => 1.27.0
zlib => 1.3
brotli => E16781312/D16781312
mutex_timedlock => enabled
pthread_barrier => enabled
mysqlnd => enabled
async_redis => enabled
coroutine_pgsql => enabled
coroutine_sqlite => enabled

Directive => Local Value => Master Value
swoole.enable_coroutine => On => On
swoole.enable_library => On => On
swoole.enable_fiber_mock => Off => Off
swoole.enable_preemptive_scheduler => Off => Off
swoole.display_errors => On => On
swoole.use_shortname => On => On
swoole.unixsock_buffer_size => 8388608 => 8388608
```

as installed from https://pkgs.alpinelinux.org/packages?name=php82-pecl-swoole&branch=v3.19&repo=community&arch=x86_64&maintainer=

5. What is your machine environment used (show your `uname -a` & `php -v` & `gcc -v`) ?

```sh
$ uname -a
[...]

PR #5484, opened 2024-09-16
[5.1] Fix pgsql header version error

NathanFreeman:


jhm-ciberman:
@NathanFreeman Sorry my ignorance, does this have something to do with this?

https://github.com/swoole/swoole-src/issues/5313#issuecomment-2349906699

PR #5554, opened 2024-11-03
[6.0]support coroutine lock feature

NathanFreeman:


src/lock/coroutine_lock.cc
```diff
@@ -0,0 +1,118 @@
...
+#else
+    return 0;
+#endif
+}
+
+int CoroutineLock::lock_impl(bool blocking) {
+    Coroutine *current_coroutine = Coroutine::get_current();
+    if (current_coroutine == nullptr) {
+        swoole_warning("The coroutine lock can only be used in a coroutine environment");
+        return 1;
```
matyhtf, line 81: Should return an error code

src/lock/coroutine_lock.cc
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    int result = 0;
+#ifdef HAVE_IOURING_FUTEX
+    if (!sw_atomic_cmp_set(value, 0, 1)) {
+        if (!blocking) {
+            return 1;
+        }
+
+        result = Iouring::futex_wait((uint32_t *) value);
```
matyhtf, line 95: No check result is given. If the wait fails, it means that the lock ownership is not obtained.

matyhtf (changes requested):

Issue #5577, opened 2024-11-21
[6.0.0RC1] build failure


remicollet:
```
[...]

PR #5578, opened 2024-11-21
[6.0]Fix bug #5577

NathanFreeman:


remicollet:
Sorry, I miss your PR while working on a fix on my side

BTW using "bc"  may raise issue if command not available, (so need to check if available...)
adding some message for debug (see PR #5579) may be useful

PR #5622, opened 2024-12-16
swoole_substr_decoding: improved parameters validation

Appla:
1. return null on error to match with built-in `json_decode` in PHP < 8.0
2. Check input offset and length to avoid out of bounds reading

 [PHP 8 style alternative](https://github.com/Appla/swoole-src/tree/sw_var_codec_php8)


matyhtf:
Thank you for your commit. The checks for the offset limits are now more rigorous than before.

Issue #5652, opened 2025-01-04
error: ‘php_swoole_serialize’ was not declared in this scope


Ser5:
Please answer these questions before submitting your issue.

1. What did you do? If possible, provide a simple script for reproducing the error.
https://github.com/swoole/swoole-src
Ran this command:
```make && make install```

2. What did you expect to see?
Some success outcome.

3. What did you see instead?
```
/bin/bash /home/ser5/swoole-src/libtool --mode=compile g++ -I. -I/home/ser5/swoole-src -I/home/ser5/swoole-src/include -I/home/ser5/swoole-src/main -I/home/ser5/swoole-src -I/usr/include/php/20210902 -I/usr/include/php/20210902/main -I/usr/include/php/20210902/TSRM -I/usr/include/php/20210902/Zend -I/usr/include/php/20210902/ext -I/usr/include/php/20210902/ext/date/lib -I/home/ser5/swoole-src -I/home/ser5/swoole-src/include -I/home/ser5/swoole-src/ext-src -I/home/ser5/swoole-src/thirdparty  -DHAVE_CONFIG_H  -g -O2 -Wall -Wno-unused-function -Wno-deprecated -Wno-deprecated-declarations -std=c++11    -DENABLE_PHP_SWOOLE -DZEND_COMPILE_DL_EXT=1 -c /home/ser5/swoole-src/ext-src/swoole_server.cc -o ext-src/swoole_server.lo  -MMD -MF ext-src/swoole_server.dep -MT ext-src/swoole_server.lo
[...]

PR #5655, opened 2025-01-06
README.md: elaboration for cloning from git

Ser5:
Added warning about cloning master branch - that using it is not recommended because it may be in development. Changed clone command so it indicates a specific production ready version.


Ser5:
In regard of https://github.com/swoole/swoole-src/issues/5652

leocavalcante (approved):

PR #5847, opened 2025-08-25
add SWOOLE_ODBC_LIBS variable to config.m4

henderkes:
Background: when statically compiling libodbc.a with libiconv.a, linking against only libodbc.a leads to unresolved external symbols to libiconv_* functions.

Here we add an optional SWOOLE_ODBC_LIBC variable if the odbc library needs more than to link against a single library.

Now `./configure --enable-swoole-odbc=unixodbc,/path/to/unixodbc SWOOLE_ODBC_LIBS="-lodbc -liconv"` succeeds.

Feel free to change this to a better way if you can think of one! https://github.com/crazywhalecc/static-php-cli/pull/864

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -655,7 +655,7 @@ EOF
       EXTRA_CFLAGS="$EXTRA_CFLAGS -I$pdo_cv_inc_path $PDO_ODBC_INCLUDE"
 
       dnl Check first for an ODBC 1.0 function to assert that the libraries work
-      save_LIBS="$LIBS"
+      SAVE_LIBS="$LIBS"
       LIBS="$LIBS $PDO_ODBC_LDFLAGS"
 
       AC_LINK_IFELSE(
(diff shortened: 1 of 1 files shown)


config.m4
```diff
@@ -641,31 +641,62 @@ EOF
...
 
       dnl Check first for an ODBC 1.0 function to assert that the libraries work
-      PHP_CHECK_LIBRARY($pdo_odbc_def_lib, SQLBindCol,
-      [
-        dnl And now check for an ODBC 3.0 function to assert that they are *good*
-        dnl libraries.
-        PHP_CHECK_LIBRARY($pdo_odbc_def_lib, SQLAllocHandle,
-        [], [
-          AC_MSG_ERROR([
+      save_LIBS="$LIBS"
```
matyhtf, line 658: The `save_LIBS` should be changed to uppercase

matyhtf (changes requested):

PR #5862, opened 2025-09-24
Fix type cast for lockwait

Appla:
The current implementation of lockwait discards the decimal portion of the input. For example, lockwait(1.2) is incorrectly converted to 1 * 1000 = 1000 milliseconds, instead of the expected 1.2 * 1000 = 1200 milliseconds.

This MR is also applicable to the master branch.


matyhtf:
@Appla There is a bug here. Your code is correct and has been merged

PR #5864, opened 2025-09-24
lock: support auto dtor options and robust for pthread_mutex

Appla:
After this PR, `Swoole\Lock` of type MUTEX can be safely shared across processes, even if a lock holder exits or is killed unexpectedly. The lock will automatically recover via robust mutex handling.

The newly introduced flags may need to be reconsidered and exported as constants for PHP users.

### DIFF
--- a/ext-src/stubs/php_swoole_lock.stub.php
+++ b/ext-src/stubs/php_swoole_lock.stub.php
@@ -1,7 +1,7 @@
 <?php
 namespace Swoole {
     class Lock {
-        public function __construct(int $type = SWOOLE_MUTEX, string $filename = '') {}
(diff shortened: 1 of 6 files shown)


matyhtf:
Version 5.1 is a branch of feature locking, which no longer accepts API changes and needs to be merged into the master branch

Appla:
> Version 5.1 is a branch of feature locking, which no longer accepts API changes and needs to be merged into the master branch

Changed the base branch to master now.

matyhtf:
The `ROBUST Mutex` is an experimental property. It is considered to replace the spin lock in table to solve the problem that the lock is not released when the process crashes and exits. It is no longer needed. It has been deleted and relevant codes will not be modified

PR #5865, opened 2025-09-26
fix: ensure timespec.tv_nsec < 1_000_000_000 to meet POSIX constraints

Appla:
rt


matyhtf:
`time ->tv_nsec` will not exceed 1 billion. The maximum value of `add_msec % 1000` is 999, multiplied by 1 million, and is always less than 1 billion.

PR #5866, opened 2025-09-26
Add timed lock support to RWLock

Appla:
This PR introduces *timeout-based locking support* for the `RWLock` type in `Swoole\\Lock`, aligning its behavior with `Mutex` based.

Changes introduced:

- `lock_read()` → `lock_read(?float $timeout = null)`: Enables acquiring a read lock with an optional timeout.
- `lockwait(float $timeout)`: Extended to support `RWLock` in addition to `Mutex`.

These PHP-level API changes introduce new semantics around lock acquisition timing and may warrant further discussion regarding coroutine compatibility and expected behavior.

### DIFF
--- a/ext-src/stubs/php_swoole_lock.stub.php
+++ b/ext-src/stubs/php_swoole_lock.stub.php
@@ -6,7 +6,7 @@ namespace Swoole {
         public function lock(): bool {}
         public function locakwait(float $timeout = 1.0): bool {}
         public function trylock(): bool {}
-        public function lock_read(): bool {}
+        public function lock_read(?float $timeout = null): bool {}
         public function trylock_read(): bool {}
         public function unlock(): bool {}
     }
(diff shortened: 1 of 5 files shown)


include/swoole_lock.h
```diff
@@ -103,6 +103,8 @@ class SpinLock final : public Lock {
     int unlock() override;
     int trylock_rd() override;
     int trylock() override;
+    int lock_wait(int timeout_msec);
```
matyhtf, line 106: It should be a member method of RWLock not SpinLock

ext-src/stubs/php_swoole_lock.stub.php
```diff
@@ -6,7 +6,7 @@ public function __destruct() {}
         public function lock(): bool {}
         public function locakwait(float $timeout = 1.0): bool {}
         public function trylock(): bool {}
-        public function lock_read(): bool {}
+        public function lock_read(?float $timeout = null): bool {}
```
matyhtf, line 9: Do not modify lock_rd() method, but modify lock_wait(), add an int type parameter, and specify whether the lock type is wrlock or rdlock.

matyhtf (changes requested):

Appla:
updated as suggestions with additional stub definition typo fixed

PR #5907, opened 2025-11-01
Correcting an error in the port.cc file

XDRiVE888:
Fixed a variable comparison error that caused multipart/form-data data to be incorrectly processed in an asynchronous HTTP server. Before this fix, large uploaded files accumulated in RAM before being written to disk and were then saved to disk after the download was complete, potentially causing RAM overflows and other potential errors. After this fix, data from files uploaded to the server is written to disk as it arrives.

### DIFF
--- a/src/server/port.cc
+++ b/src/server/port.cc
@@ -564,7 +564,7 @@ _parse:
                          request->chunked);
         if (request->form_data_) {
             if (serv->upload_max_filesize > 0 &&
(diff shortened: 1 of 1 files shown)


XDRiVE888:
It seems I was hasty and misconfigured the upload_max_filesize and package_max_length settings in my example, setting the latter to be greater than the former, and forgot to run general testing to verify the patch. My apologies.

PR #5947, opened 2025-12-03
Fix curl memory leak

NathanFreeman:


ctyytc:
What is the version range affected by this bug?

NathanFreeman:
This bug only affects Swoole 6.1.3; all other versions remain unaffected. @ctyytc

ctyytc:
Ok, understand.> This bug only affects Swoole 6.1.3; all other versions remain unaffected. @ctyytc

PR #6018, opened 2026-03-05
Feature: Export zend_autoload original handler function pointer

cheplv:
Allow modules that depends on swoole module, hook zend_autoload function.

### DIFF
--- a/ext-src/php_swoole_private.h
+++ b/ext-src/php_swoole_private.h
@@ -190,6 +190,9 @@ extern zend_class_entry *swoole_error_ce;
 extern zend_class_entry *swoole_resolve_context_ce;
 extern zend_object_handlers swoole_resolve_context_handlers;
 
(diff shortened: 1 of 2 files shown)


iconfarvie-coder:
Management.alpha@icloud.com

cheplv:
@NathanFreeman Looks like problem with arm build?

NathanFreeman:
Since the dependency packages for the Oracle SDK under the ARM architecture can no longer be downloaded, all tests and compilations related to ARM64 will result in errors. Until a solution is found, I will consider temporarily disabling the Oracle database tests for ARM64.

cheplv:
@NathanFreeman May be use this repo as temporary solution: https://github.com/yalelibrary/instantclient-binaries

cheplv:
@NathanFreeman Due oracle arm build is fixed, think need to rerun build again?

PR #6019, opened 2026-03-07
docs: add `sudo` privileges to examples

santysisi:
Add `sudo` to the build example (`make install` step) in the documentation.

The docs recommend using `sudo` privileges. This change also keeps the instructions consistent with other sections of the `Markdown` documentation where `sudo` is already specified for the installation step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,7 +581,7 @@ Compile and install at the source folder:
 ```shell
 phpize && \
 ./configure && \
-make && make install
+make && sudo make install
 ```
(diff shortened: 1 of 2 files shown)


santysisi:
I noticed some tests are failing. Since it’s late here, I’ll investigate those tomorrow.

Also, I’m not sure if I am targeting the correct branch. Please let me know if you would prefer me to rebase onto 6.1 instead.

PR #6025, opened 2026-03-11
Compile against PHP 8.6

arshidkv12:


NathanFreeman:
Thank you very much for your submission! However, PHP 8.5 has only been released for less than six months, and PHP 8.6 still has many uncertain changes. Therefore, Swoole will not be adapting to PHP 8.6 for the time being. I have reviewed your modifications, which primarily involve function name changes specific to PHP 8.6. Such adjustments would prevent normal usage on PHP 8.5 and earlier versions.

arshidkv12:
Thank you
Please check the new changes.

matyhtf:
Support for 8.6 will not be considered in the short term. There is a big difference from the master, so this PR will be temporarily closed.

PR #6027, opened 2026-03-13
Feature: Export zend_autoload original handler function pointer

cheplv:
Allow modules that depends on swoole module, hook zend_autoload function.


cheplv:
Pull request recreated due previous was closed on deleted 6.2 branch

PR #6042, opened 2026-04-15
Fix memory leak when server in single worker mode

cheplv:
Changes fixes memory leak when server is run in single worker mode and clears references of network address maps


src/server/master.cc
```diff
@@ -761,9 +761,14 @@ Server::~Server() {
     if (!is_shutdown() && getpid() == gs->master_pid) {
         destroy();
     }
+
     for (auto port : ports) {
         delete port;
     }
+
+    local_addr_v4_map.clear();
+    local_addr_v6_map.clear();
```
matyhtf: This code is redundant; C++ objects are automatically cleaned up when they are destroyed.

matyhtf (changes requested):

src/server/master.cc
cheplv: @matyhtf Currently valgrind see memleaks on server object, it doesn't destructed.
Will trace deep tomorrow and add more changes

cheplv:
@matyhtf Looks like some of "macOS" checks not working correctly?

cheplv:
@matyhtf Can you please review?

ext-src/swoole_server.cc
```diff
@@ -233,7 +233,7 @@ static void server_free_object(zend_object *object) {
     delete property;
 
     zend_object_std_dtor(object);
-    if (serv && serv->is_master()) {
+    if (serv && (serv->is_master() || serv->is_single_worker()) ) {
```
NathanFreeman, line 236: Thank you for your work. Do you have a detailed memory leak report available? We need to note that in SWOOLE_THREAD mode, the master thread is always present, even if the number of worker threads is set to 1.

PR #6076, opened 2026-06-06
Fix: Remove duplicate key add for request server array

cheplv:
Remove duplicate key add for request time, request time float  and server protocol from server array.

Http request parser already define this vars.
In debug build it throws assert when key add to server array with existing key.

https://github.com/swoole/swoole-src/blob/master/ext-src/swoole_http_request.cc#L473

### DIFF
--- a/ext-src/swoole_http_server_coro.cc
+++ b/ext-src/swoole_http_server_coro.cc
@@ -628,9 +628,6 @@ static PHP_METHOD(swoole_http_server_coro, onAccept) {
         zval *zserver = ctx->request.zserver;
         HashTable *ht = Z_ARRVAL_P(zserver);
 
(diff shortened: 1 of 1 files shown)


cheplv:
@NathanFreeman Good day,

Please check this PR. Fixes debug build issue.

NathanFreeman:
I don't see any issues with this PR, but it still needs @matyhtf  to review it. He's been a bit busy lately.

cheplv:
Ok. Thx.

cheplv:
@matyhtf Can you please review?

matyhtf:
These headers have been set in the http_request_on_headers_complete function. They are indeed duplicate codes and can be deleted. Merged

PR #6085, opened 2026-06-24
Fix io_uring kernel version check to correctly handle major version > 6

### LINKED ISSUE #6083: Swoole io_uring kernel version gate rejects Linux 7.0 as lower than 6.7
# Swoole io_uring kernel version gate rejects Linux 7.0 as lower than 6.7

## Summary

Swoole 6.2.1 with io_uring enabled fails to initialize on a Linux 7.0 kernel because the runtime kernel-version guard checks `major >= 6 && minor >= 7`.

For Linux `7.0`, this evaluates to false even though `7.0` is newer than `6.7`. The same pattern also appears in the `ftruncate()` gate for `6.9`.

## Environment

- Swoole: `6.2.1`
- PHP: `8.5.7`
- Alpine: `3.24.1`
- Package: `php85-pecl-swoole 6.2.1-r0`
- liburing: `2.14-r0`
- Kernel: `Linux 7.0.0-22-generic`
- Docker: tested with `--security-opt seccomp=unconfined`

`php85 --ri swoole` reports:

```text
Version => 6.2.1
io_uring => enabled
```

## Reproduction

Note: default Docker seccomp blocks `io_uring_setup()` with `EPERM`, so this repro intentionally uses `--security-opt seccomp=unconfined` to isolate the Swoole version check.

```sh
docker run --rm --security-opt seccomp=unconfined alpine:latest sh -ec '
  apk add --no-cache php85 php85-pecl-swoole >/tmp/apk-install.log
  uname -r
  php85 -v
  php85 --ri swoole | grep -E "^(Version|io_uring)"
  php85 -r '"'"'
    swoole_async_set([
      "iouring_entries" => 128,
      "iouring_workers" => 4,
      "iouring_flag" => SWOOLE_IOURING_DEFAULT,
    ]);
    Swoole\Runtime::enableCoroutine(SWOOLE_HOOK_FILE);
    Co\run(function () {
      file_put_contents("/tmp/swoole-iouring-test.txt", "abc");
      echo file_get_contents("/tmp/swoole-iouring-test.txt"), PHP_EOL;
    });
  '"'"'
'
```

## Actual result

```text
7.0.0-22-generic
PHP 8.5.7 (cli) ...
Version => 6.2.1
io_uring => enabled
[ERROR] Iouring::Iouring(): The Iouring::futex_wait()/Iouring::futex_wakeup() requires `6.7` or higher Linux kernel
```

## Expected result

Linux 7.0 should satisfy checks for features requiring Linux 6.7 or 6.9, and Swoole should proceed to initialize io_uring.

## Suspected cause

In `src/coroutine/iouring.cc`, the current guards compare major and minor independently:

```cpp
if (!(major >= 6 && minor >= 7)) {
    swoole_error("The Iouring::futex_wait()/Iouring::futex_wakeup() requires `6.7` or higher Linux kernel");
}

if (!(major >= 6 && minor >= 9)) {
    swoole_error("The Iouring::ftruncate() requires `6.9` or higher Linux kernel");
}
```

For kernel `7.0`, `major >= 6` is true but `minor >= 7` is false, so the guard rejects it.

A version comparison helper would avoid this:

```cpp
static bool kernel_version_at_least(int major, int minor, int required_major, int required_minor) {
    return major > required_major || (major == required_major && minor >= required_minor);
}
```

Then:

```cpp
if (!kernel_version_at_least(major, minor, 6, 7)) {
    ...
}

if (!kernel_version_at_least(major, minor, 6, 9)) {
    ...
}
```

## Additional verification

As a sanity check, I used a temporary `LD_PRELOAD` shim that only changes `uname(2)`'s release string from `7.0.0-22-generic` to `6.9.0-test`. With the same PHP/Swoole binary and Docker `seccomp=unconfined`, coroutine file/socket io_uring tests pass.

`strace` then shows real io_uring syscalls:

```text
io_uring_setup(128, ...) = 5
io_uring_register(5, IORING_REGISTER_IOWQ_MAX_WORKERS, ...) = 0
io_uring_enter(5, 1, 0, 0, NULL, 8) = 1
```

lazerg:
The condition `major >= 6 && minor >= N` incorrectly rejects kernels where the major version exceeds 6. For example, Linux 7.0 evaluates as `7 >= 6 && 0 >= 7` → `false`, triggering the error even though 7.0 is newer than 6.7.

The fix uses the standard version comparison: `(major > 6) || (major == 6 && minor >= N)`, applied to both the futex (≥ 6.7) and ftruncate (≥ 6.9) guards.

Fixes #6083

### DIFF
--- a/src/coroutine/iouring.cc
+++ b/src/coroutine/iouring.cc
@@ -116,13 +116,13 @@ Iouring::Iouring(Reactor *_reactor) {
     parse_kernel_version(SwooleG.uname.release, &major, &minor);
 
 #ifdef HAVE_IOURING_FUTEX
-    if (!(major >= 6 && minor >= 7)) {
+    if (!((major > 6) || (major == 6 && minor >= 7))) {
         swoole_error("The Iouring::futex_wait()/Iouring::futex_wakeup() requires `6.7` or higher Linux kernel");
     }
 #endif
 
 #ifdef HAVE_IOURING_FTRUNCATE
-    if (!(major >= 6 && minor >= 9)) {
+    if (!((major > 6) || (major == 6 && minor >= 9))) {
         swoole_error("The Iouring::ftruncate() requires `6.9` or higher Linux kernel");
     }
 #endif


lazerg:
@matyhtf could you take a look when you have a chance? This is a small two-line logic fix for the kernel version gate in io_uring.

matyhtf:
It has been reviewed that the old code cannot support Linux 7 or above, and the code change is correct.

PR #6111, opened 2026-07-21
Fix concurrent SSH2 authentication state and redact secrets

binaryfire:
## Summary

- move keyboard-interactive passwords from process-global state to the owning SSH2 session
- clear borrowed authentication state before success or password fallback
- redact passwords, in-memory private keys, and passphrases from exception traces

## Why

The keyboard-interactive callback previously read a process-global password pointer. Two coroutines authenticating separate sessions could overwrite that pointer and submit one session's password to the other server.

SSH2's secret-bearing authentication parameters also lacked `SensitiveParameter` metadata, so PHP exception traces could retain their plaintext values when argument capture was enabled.

## Implementation

The in-flight keyboard-interactive password now lives in libssh2's per-session abstract data. The callback initializes every response slot, copies the first response with the allocator registered for that session, and the caller clears both borrowed fields before returning or falling back.

The SSH2 stub marks the exact secret inputs. Since this integration uses a hand-written function table, MINIT registers the corresponding parameter attributes explicitly. A fixture-independent PHPT verifies the reflection metadata and end-to-end trace redaction.

The existing OpenSSH fixture disables keyboard-interactive authentication, so this does not add a probabilistic concurrency test or PAM-specific fixture machinery. Existing password and public-key authentication coverage verifies compatibility around the changed path.

### DIFF
--- a/ext-src/php_swoole_ssh2.h
+++ b/ext-src/php_swoole_ssh2.h
@@ -15,6 +15,11 @@ typedef struct _php_ssh2_session_data {
     zval *disconnect_cb;
 
     SocketImpl *socket;
+
+    /* Keyboard-interactive password for the in-flight auth attempt. This lives
+     * on the session so concurrent coroutines cannot overwrite callback state. */
+    const char *kbd_password;
+    size_t kbd_password_len;
 } php_ssh2_session_data;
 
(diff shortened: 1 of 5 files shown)


binaryfire:
Superseded by https://github.com/swoole/swoole-src/pull/6110

PR #6113, opened 2026-07-21
Fix SSH2 resource ownership and stream error handling

binaryfire:
## Problem

The bundled SSH2 extension has one connected set of resource-lifetime and stream-contract bugs:

- direct SFTP wrapper connections do not consistently register and release their session and SFTP resources;
- one-shot wrapper operations leak temporary references;
- child resources can call libssh2 after their parent session has been destroyed;
- channel reads, writes, EOF, blocking mode, and timeouts do not follow PHP stream semantics; and
- SFTP file and directory errors can be reported as clean EOF.

These paths share the same ownership graph and stream implementations. Splitting them across pull requests would leave unsafe intermediate states.

## Change

The series contains five implementation commits plus one test-isolation follow-up:

1. Initialize every field on SFTP directory handles before exposing the stream.
2. Make direct wrapper session/SFTP ownership transactional and release temporary wrapper resources.
3. Guard child operations and destructors with their parent resource lifetime.
4. Correct channel blocking, nonblocking backpressure, operation-local timeouts, error returns, and EOF metadata.
5. Keep SFTP file and directory failures distinct from clean completion.
6. Keep the authentication-failure regression below OpenSSH 9.8+'s default per-source penalty floor without weakening its descriptor-leak assertion.

Channel timeouts are passed directly to the owning socket poll. They no longer mutate shared session socket state, so a timeout on one channel cannot affect another channel or a later SFTP operation on the same SSH session.

## Tests

New PHPT coverage exercises:

- SFTP directory enumeration;
- direct wrapper authentication, resource registration, and reference release;
- child-resource behavior after parent-session closure;
- channel blocking mode, nonblocking backpressure, transport errors, EOF, and sub-second timeout isolation; and
- SFTP file and directory error handling.

Each implementation commit was built and tested at its own boundary. The complete added SSH2 test set also passes against the final extension build.

## Review notes

The affected ownership paths were compared with PECL ssh2 1.5.0. This keeps the established Zend resource model where it applies, while retaining Swoole's coroutine-aware I/O and adding the Swoole-specific channel timeout and nonblocking behavior.

The parent-guard commit also changes `(data->refcount)--` to `(*(data->refcount))--` in the allocation-failure cleanup for `ssh2_fetch_stream()`. That branch cannot be forced through a normal PHPT without a production test hook, but the current expression advances the pointer instead of decrementing the referenced count.

### DIFF
--- a/ext-src/php_swoole_ssh2.h
+++ b/ext-src/php_swoole_ssh2.h
@@ -31,11 +31,6 @@ static inline SocketImpl *ssh2_get_socket(LIBSSH2_SESSION *session) {
     return (*session_data)->socket;
 }
 
-static inline void ssh2_set_socket_timeout(LIBSSH2_SESSION *session, int timeout_ms) {
-    auto sock = ssh2_get_socket(session);
-    sock->set_timeout(timeout_ms / 1000, SW_TIMEOUT_ALL);
-}
-
 class ResourceGuard {
     zval zres_;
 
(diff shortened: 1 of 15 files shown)


matyhtf:
GOTO https://github.com/swoole/swoole-src/pull/6122

PR #6117, opened 2026-07-24
Fix trailers after streamed HTTP/2 responses

binaryfire:
A streamed HTTP/2 response ending without a final body currently sends an empty DATA frame with END_STREAM before its trailer HEADERS. Clients receive the body but not the trailers, and the server emits HEADERS for a closed stream.

This keeps the empty DATA terminator for trailerless streams and lets trailer HEADERS terminate responses that have trailers.

The regression test covers streamed trailers alongside the existing trailer and trailerless completion paths on one connection.

### DIFF
--- a/ext-src/swoole_http2_server.cc
+++ b/ext-src/swoole_http2_server.cc
@@ -720,11 +720,15 @@ static bool http2_server_respond(HttpContext *ctx, const String *body) {
 #endif
 
     SW_LOOP {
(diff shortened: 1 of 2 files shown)


NathanFreeman (approved):

matyhtf:
GOTO : https://github.com/swoole/swoole-src/pull/6124

PR #6118, opened 2026-07-24
Fix HTTP/2 session teardown after fatal requests

binaryfire:
This fixes a worker crash during shutdown when a fatal request leaves an HTTP/2 session alive.

Remaining sessions are now released during request shutdown, before Zend tears down the allocator used by nghttp2. Cleanup preserves the different context ownership rules for coroutine and async sessions.

The regression test covers the fatal request path and verifies that the replacement worker remains healthy.

### DIFF
--- a/ext-src/swoole_http2_server.cc
+++ b/ext-src/swoole_http2_server.cc
@@ -1297,7 +1297,6 @@ int swoole_http2_server_onReceive(Server *serv, Connection *conn, RecvData *req)
         client->default_ctx->onBeforeRequest = http2_server_onBeforeRequest;
         client->max_body_size = serv->get_package_max_length(conn);
         client->handle = http2_server_onRequest;
-        http2_sessions.emplace(session_id, client);
     } else {
         client = iter->second;
     }
@@ -1331,7 +1330,15 @@ void swoole_http2_server_session_free(SessionId session_id) {
     if (iter == http2_sessions.end()) {
         return;
     }
-    /* default_ctx does not blong to session object */
-    iter->second->default_ctx = nullptr;
+    if (iter->second->is_coro) {
+        // A coroutine session borrows default_ctx from its request/response objects; an async session owns it.
+        iter->second->default_ctx = nullptr;
+    }
     http2_sessions.erase(iter);
+}
+
+void swoole_http2_server_release_sessions() {
+    while (!http2_sessions.empty()) {
+        swoole_http2_server_session_free(http2_sessions.begin()->first);
+    }
 }
--- a/ext-src/php_swoole_http_server.h
+++ b/ext-src/php_swoole_http_server.h
@@ -40,6 +40,7 @@ int swoole_http2_server_onReceive(swoole::Server *serv, swoole::Connection *conn
 
 std::shared_ptr<swoole::http2::Session> swoole_http2_server_session_new(swoole::SessionId fd);
 void swoole_http2_server_session_free(swoole::SessionId fd);
+void swoole_http2_server_release_sessions();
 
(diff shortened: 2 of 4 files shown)


NathanFreeman:
I took a look at this unit test. When an exception is not caught, the process exiting is normal behavior, and resources will also be released at the same time.

ext-src/swoole_http2_server.cc
```diff
@@ -1297,7 +1297,6 @@ int swoole_http2_server_onReceive(Server *serv, Connection *conn, RecvData *req)
         client->default_ctx->onBeforeRequest = http2_server_onBeforeRequest;
         client->max_body_size = serv->get_package_max_length(conn);
         client->handle = http2_server_onRequest;
-        http2_sessions.emplace(session_id, client);
```
NathanFreeman, line 1300: May I ask what was the consideration behind deleting this?

ext-src/swoole_http2_server.cc
```diff
@@ -1331,7 +1330,15 @@ void swoole_http2_server_session_free(SessionId session_id) {
     if (iter == http2_sessions.end()) {
         return;
     }
-    /* default_ctx does not blong to session object */
-    iter->second->default_ctx = nullptr;
+    if (iter->second->is_coro) {
+        // A coroutine session borrows default_ctx from its request/response objects; an async session owns it.
+        iter->second->default_ctx = nullptr;
+    }
```
NathanFreeman, line 1336: I'm a bit confused here—what scenario is this iter->second->is_coro check intended to handle?"

binaryfire:
Hi @NathanFreeman 

> I took a look at this unit test. When an exception is not caught, the process exiting is normal behavior, and resources will also be released at the same time.

You're right that the worker should exit because the exception isn't caught. This PR doesn't change that. The test allows the worker to exit. It just checks that the exit wasn't caused by a signal. On master, onWorkerError reports signal 11 (SIGSEGV). With this PR, it reports signal 0.

After the fatal error, the HTTP/2 session remains in the thread-local map. The nghttp2 objects use the Zend allocator configured by `php_nghttp2_mem()`. The session is later destroyed after that allocator shuts down, so the nghttp2 cleanup crashes. This PR destroys the sessions earlier, while the allocator is still available. The worker still exits because of the PHP fatal error, but it no longer segfaults.

ext-src/swoole_http2_server.cc
binaryfire: `swoole_http2_server_session_new()` already adds the session to http2_sessions.The second emplace() uses the same key. 

`std::unordered_map::emplace()` doesn't replace an existing value, so this call doesn't do anything. I removed it because the duplicate insertion was confusing, but it's not required for this fix, so I can restore it if you prefer.

ext-src/swoole_http2_server.cc
binaryfire: `default_ctx` is handled differently by the two server types. The async server creates it with new `HttpContext()`, so the session owns it and has to delete it. The coroutine server assigns an existing context. That context is deleted by the coroutine request cleanup so the session has to clear its pointer and not delete it again.

Before this PR, `default_ctx` was always cleared. That prevents a double free in the coroutine server, but it leaks the context created by the async server. In my test, the worker RSS grew by about 1 KB for every closed HTTP/2 connection.

The `is_coro` check keeps the correct behavior for both cases.

NathanFreeman (approved):

NathanFreeman:
I see what you mean. When the PHP process exits, nghttp2 needs to release its resources, but by that time the PHP environment has already been corrupted, so efree cannot be used to free the resources.

PR #6121, opened 2026-07-24
Add optional file deletion after sendfile

binaryfire:
Frameworks often serve generated or temporary files that should be deleted after the response finishes using them. In `SWOOLE_PROCESS` mode, `sendfile()` queues a pathname and returns before the reactor opens it. Swoole provides no completion signal, so a framework cannot know when it can safely remove the file. It must stream the file through PHP or delay cleanup. In `SWOOLE_BASE` mode the file is opened before `sendfile()` returns for the connection handled by that worker; the same gap exists when a send is forwarded to another worker or thread.

This adds an optional `delete_file` argument to `Response::sendfile()` and `Server::sendfile()`. When enabled, Swoole takes responsibility for deletion and removes the source file after the native transfer finishes or terminates, including send errors and client disconnects. `sendfile()` can still return `false` after that responsibility has moved to Swoole, so a `false` return does not mean the file still exists. Existing behavior remains unchanged by default.

This does not prevent application code from independently deleting or replacing the path. It gives callers that request cleanup a safe way to retain native `sendfile()`.

### DIFF
--- a/ext-src/php_swoole_http.h
+++ b/ext-src/php_swoole_http.h
@@ -202,7 +202,7 @@ struct Context {
     // The `private_data_2` pointer is used to save callback function
     void *private_data_2;
     bool (*send)(Context *ctx, const char *data, size_t length);
-    bool (*sendfile)(Context *ctx, zend_string *file, off_t offset, size_t length);
+    bool (*sendfile)(Context *ctx, zend_string *file, off_t offset, size_t length, bool &delete_file);
     bool (*close)(Context *ctx);
     bool (*onBeforeRequest)(Context *ctx);
     void (*onAfterResponse)(Context *ctx);
@@ -225,7 +225,7 @@ struct Context {
     HTTP_API bool set_header(const char *, size_t, const std::string &, bool);
     HTTP_API void end(zend_string *sdata, zval *return_value);
     HTTP_API void write(zend_string *sdata, zval *return_value);
-    HTTP_API bool send_file(zend_string *file, off_t offset, size_t length);
+    HTTP_API bool send_file(zend_string *file, off_t offset, size_t length, bool delete_file);
 
     String *get_write_buffer();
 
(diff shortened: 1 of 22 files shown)


NathanFreeman:
Thank you for your work, but I believe this fix only addresses the symptoms rather than the root cause. Even with this restriction in place, there is still no way to prevent users from deleting files after they have been sent.

binaryfire:
@NathanFreeman I have updated the title and description to make the intended scope clearer.

You are right that this option does not prevent application code from deleting or replacing the path after `sendfile()` returns. It is not intended to solve that general race.

The gap we are addressing is cleanup of framework-owned temporary files. The Swoole coroutine server finishes sending the file before `sendfile()` returns, and `SWOOLE_BASE` mode opens it before returning for a locally handled connection. Other file-serving APIs commonly either finish with the source before returning or provide a completion callback. The Swoole process-mode path does neither, so a framework cannot know when it can safely delete the source file.

[Hypervel currently handles Symfony’s `deleteFileAfterSend()` setting by streaming the file through PHP](https://github.com/hypervel/components/blob/70b69d20ba089f50dd33db31905e126dfd442a2b/src/http-server/src/ResponseBridge.php#L148-L196) instead of using native `sendfile()`. That fallback reads and writes the entire file through PHP in 16 KiB chunks by default, adding work and inter-process copying to every transfer. It also removes `Content-Length` because `Response::write()` uses chunked transfer. This option would let any framework transfer cleanup to Swoole while retaining native `sendfile()`.

A general fix would need the worker/reactor path to establish ownership of an open file, for example by passing an open descriptor instead of only the pathname. That is a much larger change to the worker/reactor message bus. This option is the small, explicit alternative: the caller asks Swoole to manage cleanup, and Swoole deletes the file when the transfer ends.

Would you accept it as that narrower feature, or would you prefer a different shape for it?

PR #6126, opened 2026-07-29
Add atomic row operations to Swoole Table

binaryfire:
## Summary

Add five atomic row operations to `Swoole\Table`:

```php
$table->add(string $key, array $values): bool;
$table->update(string $key, array $values): bool;
$table->cmpset(string $key, array $expected, array $values): bool;
$table->cmpdel(string $key, array $expected): bool;
$table->getdel(string $key, ?string $field = null): array|false|string|float|int;
```

Each method evaluates its condition and applies its mutation under one existing Table bucket lock.

## Motivation

Absent-only insertion currently requires separate calls:

```php
if (!$table->exists($key)) {
    $table->set($key, $values);
}
```

The bucket lock is released between those calls. Another process can change the row between the check and mutation, so callers need a second shared lock to make the operation safe. The same race affects read-compare-update and read-compare-delete sequences.

The new methods expose the transactions that the native Table lock can already serialize:

- `add()` inserts only when the physical row is absent.
- `update()` updates only when the row is present.
- `cmpset()` updates only when selected columns still match.
- `cmpdel()` deletes only when selected columns still match.
- `getdel()` returns and deletes a row atomically.

The names follow existing Swoole APIs such as `Swoole\Thread\Map::add()`, `Swoole\Thread\Map::update()`, and `Swoole\Atomic::cmpset()`.

## Implementation

Each operation hashes once, acquires one bucket lock, and traverses the collision chain once.

Comparisons use the stored representation. Integers and floats compare their exact native bytes. Strings compare their logical length and bytes. Expected strings are never truncated. This makes values read from Table suitable for later compare-and-set operations, including signed zero and NaN values.

Expected values are strict: they must contain at least one declared column, and any invalid key rejects the complete condition. Write values retain `set()` compatibility by ignoring numeric and unknown keys.

PHP values are converted before the shared-memory lock is acquired. This keeps callbacks, warnings, and string conversion outside the bucket spinlock. A small nesting guard prevents `destroy()` from invalidating native schema pointers during conversion.

`getdel()` copies the requested row or field while locked, deletes the row, releases the lock, and then converts the owned snapshot into a PHP value.

The existing `set()`, `get()`, and `del()` paths share the same collision lookup and deletion mechanics without changing their public behavior, flags, or statistics.

This change is limited to row existence and equality conditions. It does not add callbacks, predicates, TTL handling, aliases, or multi-row transactions.

## Testing

The native and PHP tests cover:

- absent-only and present-only writes;
- successful and failed comparisons;
- integer, float, binary-string, signed-zero, and NaN behavior;
- root, middle, and tail collision-chain mutations;
- allocation exhaustion and pool-slice reuse;
- operation statistics;
- full-row and field `getdel()` snapshots;
- multi-process contention and versioned compare-and-set retries;
- conversion re-entry and Table lifecycle safety.

### DIFF
--- a/core-tests/src/memory/table.cpp
+++ b/core-tests/src/memory/table.cpp
@@ -131,6 +131,40 @@ class table_t {
     }
 };
 
+static TableValue table_int(Table *table, const char *column, long value) {
+    return TableValue(table->get_column(column), &value, sizeof(value));
+}
+
+static TableValue table_float(Table *table, const char *column, double value) {
+    return TableValue(table->get_column(column), &value, sizeof(value));
+}
+
+static TableValue table_string(Table *table, const char *column, const std::string &value) {
+    return TableValue(table->get_column(column), value.data(), value.size());
+}
+
+static TableValues table_row(Table *table, const row_t &value) {
+    return {
+        table_int(table, "id", value.id),
+        table_string(table, "name", value.name),
+        table_float(table, "score", value.score),
+    };
+}
+
+static row_t table_snapshot(Table *table, const std::string &data) {
+    row_t value;
+    const auto column_id = table->get_column("id");
+    const auto column_name = table->get_column("name");
+    const auto column_score = table->get_column("score");
+
+    memcpy(&value.id, data.data() + column_id->index, sizeof(value.id));
+    memcpy(&value.score, data.data() + column_score->index, sizeof(value.score));
+    TableStringLength length;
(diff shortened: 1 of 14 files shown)


matyhtf:
Thank you for your contribution. These APIs are very useful designs.

binaryfire:
@matyhtf No problems. Swoole is a great project - I look forward to contributing more.

PR #6136, opened 2026-08-01
Fix reentrant argument conversion in Table and constructors

binaryfire:
## Summary

PHP argument conversion can invoke user code. Several Table operations and object constructors inspected native state or acquired row locks before conversion completed. Re-entry could then initialize or destroy the same object while the outer call retained stale state.

Table values are now converted before row locks are acquired. A new internal C++ `Table::set()` overload applies the prepared values while holding the lock, preserving atomic row updates without changing the PHP API. `Table::count()` also parses its argument before reading the native Table pointer.

The same ordering is applied to the Table, Lock, Coroutine Lock, Thread Lock, Thread Barrier, Process, and Server constructors. Server construction now releases native Server and ListenPort state when setup fails before the object is published.

## Testing

- core coverage for prepared Table writes
- PHPT coverage for Table conversion, constructor, and count re-entry
- constructor re-entry coverage for locks, thread primitives, Process, and Server

### DIFF
--- a/core-tests/src/memory/table.cpp
+++ b/core-tests/src/memory/table.cpp
@@ -241,6 +241,80 @@ TEST(table, conditional_writes) {
     ASSERT_EQ(table.get("php").id, 2);
 }
 
+TEST(table, values_set) {
+    table_t table(128);
+    auto ptr = table.ptr();
+
+    bool out_of_space = true;
+    ASSERT_TRUE(ptr->set("new", 3, {table_int(ptr, "id", 1)}, &out_of_space));
+    ASSERT_FALSE(out_of_space);
+    auto row = table.get("new");
+    ASSERT_EQ(row.id, 1);
+    ASSERT_EQ(row.name, "");
+    ASSERT_EQ(row.score, 0);
+    ASSERT_EQ(table.count(), 1);
+    ASSERT_EQ(ptr->insert_count, 1);
+    ASSERT_EQ(ptr->update_count, 0);
+
+    ASSERT_TRUE(ptr->set("new", 3, {table_string(ptr, "name", "updated")}, &out_of_space));
+    ASSERT_FALSE(out_of_space);
+    row = table.get("new");
+    ASSERT_EQ(row.id, 1);
+    ASSERT_EQ(row.name, "updated");
+    ASSERT_EQ(row.score, 0);
+    ASSERT_EQ(table.count(), 1);
+    ASSERT_EQ(ptr->insert_count, 1);
+    ASSERT_EQ(ptr->update_count, 1);
+
+    ASSERT_TRUE(ptr->set("empty", 5, {}));
+    ASSERT_TRUE(ptr->set("empty", 5, {}));
+    ASSERT_EQ(table.count(), 2);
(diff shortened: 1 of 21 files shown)


matyhtf (dismissed):
@-

matyhtf (changes requested):
The conversion-safety changes for `set()`, `incr()`, and `decr()` look sound, but one lifecycle path is still unsafe and should be fixed before merging.

`Swoole\\Table::__construct()` reads the native pointer before parsing `table_size` and `conflict_proportion`. Scalar parameter conversion can invoke a user error handler, for example during a lossy float-to-int conversion. That handler can re-enter `__construct()`, initialize the same object, and even create the table. When control returns to the outer constructor, it continues and overwrites the pointer with a second `Table`. The re-entrantly created table, including any allocated shared memory, then becomes unreachable, and its schema is lost.

I reproduced this by creating an uninitialized Table through reflection, installing an error handler that calls `__construct(8)` and adds a column, and then calling `__construct(8.5)`. The outer call replaces the inner table, and the column added by the handler disappears.

Please parse the constructor arguments before the first `table_get_ptr()` check, then check whether the object was initialized during conversion. In that case the outer call should report that the constructor can only be called once instead of overwriting the pointer. Please also add a PHPT regression test for this re-entrant constructor case.

The currently failing Table-unrelated CI jobs appear flaky or external; the new Table PHPTs and native Table tests themselves pass.

matyhtf:
Suggested implementation:

```cpp
PHP_METHOD(swoole_table, __construct) {
    zend_long table_size;
    double conflict_proportion = SW_TABLE_CONFLICT_PROPORTION;

    ZEND_PARSE_PARAMETERS_START_EX(ZEND_PARSE_PARAMS_THROW, 1, 2)
    Z_PARAM_LONG(table_size)
    Z_PARAM_OPTIONAL
    Z_PARAM_DOUBLE(conflict_proportion)
    ZEND_PARSE_PARAMETERS_END_EX(RETURN_FALSE);

    // Parameter conversion may execute a user error handler and re-enter
    // __construct(), so native state must only be checked after parsing.
    Table *table = table_get_ptr(ZEND_THIS);
    if (table) {
        zend_throw_error(nullptr, "Constructor of %s can only be called once", SW_Z_OBJCE_NAME_VAL_P(ZEND_THIS));
        RETURN_FALSE;
    }

    table = Table::make(table_size, static_cast<float>(conflict_proportion));
    if (table == nullptr) {
        zend_throw_exception(swoole_exception_ce, "global memory allocation failure", SW_ERROR_MALLOC_FAIL);
        RETURN_FALSE;
    }

    // Keep the existing hash-function setup and scope-exit cleanup here.

    table_set_ptr(ZEND_THIS, table);
}
```

The important part is moving both parameter parsing and conversion before the first `table_get_ptr()`. If conversion re-enters and initializes the object, the outer call will then detect that state and fail without replacing or leaking the inner `Table`.

binaryfire:
@matyhtf  Thanks, you were right. I pushed a follow-up that validates constructor arguments before checking whether the object was already initialized. That avoids leaving partially updated native state visible if argument conversion runs user code.

I also applied the same ordering to the nearby constructors that had the same shape.

PR #6141, opened 2026-08-09
Fix SSH2 resource release lifecycle and SCP failures

binaryfire:
## Summary

Swoole keeps libssh2 sessions nonblocking, but several resource destructors called their free operation once and discarded `EAGAIN`. Undrained channels, SFTP resources, listeners, and their session-owned state could therefore remain allocated. Callback containers were also retained during normal teardown and failed handshakes.

This change adds one internal release path:

- releases inside a coroutine yield until completion or timeout;
- top-level releases use a bounded synchronous wait and restore session state;
- releases on an already-owned socket or during request shutdown make one nonblocking attempt;
- final session teardown shuts down the transport, performs a bounded blocking free, and releases wrapper state only after libssh2 confirms success.

If a fatal child thread prevents cancellation from unwinding a bound reader during request shutdown, the transport descriptor is still closed without deleting the bound socket wrapper.

Ordinary SSH2 I/O gains no lock or additional synchronization. Closing an SSH2 stream inside a coroutine may now yield up to its effective timeout. A timed-out child remains session-owned until teardown. The small public-key wrapper can still remain allocated if its single-attempt shutdown returns `EAGAIN`, because libssh2 provides no ownership hook for reclaiming it safely.

The PR also reads libssh2's poll direction after the current operation returns `EAGAIN`, stops truncated SCP receives instead of spinning on early EOF, and rejects failed or incomplete local writes.

## Testing

Added focused PHPT coverage for channel, session, SFTP, listener, callback, request-shutdown, concurrency, and SCP failure paths. A Linux thread regression also verifies that fatal child-thread shutdown does not leave the SSH transport descriptor open; it fails on the previous PR head and passes with this change.

The complete SSH2 suite passes in the repository's PHP 8.4 container configured with thread support, `--enable-sockets`, and `--with-swoole-ssh2`. GitHub workflows do not provision the SSH2 fixture, so the focused container run provides this validation.


matyhtf:
Thanks for working through the SSH2 release lifecycle. Moving conversion/release waits outside unsafe ownership contexts, reading the block direction after `EAGAIN`, and covering the SCP failure cases all look useful.

There is one final teardown path that still needs to be made reliable before merging:

In `php_ssh2_session_dtor()`, the session is still restored to nonblocking mode after `ssh2_release_call()`. The code then calls `shutdown(SHUT_RDWR)`, invokes `libssh2_session_free(session)` once without checking its result, and immediately frees `session_data`.

`libssh2_session_free()` is explicitly allowed to return `LIBSSH2_ERROR_EAGAIN`. Its implementation preserves the session when an attached channel or listener cannot yet be released. If that happens here, the libssh2 session/resource graph remains allocated while its abstract pointer, callbacks, and `SocketImpl` are freed. This leaves a native leak and a dangling abstract pointer. A successful transport shutdown will usually make child cleanup local, but that is not an API guarantee, and the return value of `shutdown()` is also currently ignored.

Please make the final session free non-`EAGAIN` and only release `session_data` after confirming that the session was actually freed. One possible shape is to shut down the transport, set a bounded libssh2 timeout, switch the session to blocking mode, and then call `libssh2_session_free()`. Since a successful call invalidates the session, this final call does not need to restore the old mode. Please also cover the case where the first final free attempt would return `EAGAIN`.

There is also a small formatting issue: `git diff --check` reports trailing whitespace on the newly added `.gitignore` line for `truncated_scp_server.sh`.

The Linux unit-test failures shown in CI appear unrelated to this PR (external HTTP/proxy failures rather than SSH2 failures).

binaryfire:
@matyhtf Fixed. Final session teardown now shuts down the transport, sets a one-second libssh2 timeout, switches to blocking mode, and checks `libssh2_session_free()`. If its first internal attempt returns `EAGAIN`, libssh2 retries within that bound. Session data is released only after a successful free; otherwise it is retained and the warning includes libssh2's error.

`shutdown()` remains best effort because the checked, bounded free is the ownership boundary. If shutdown does not bring the transport down, the free either completes or reaches the bound without releasing data that libssh2 still references.

I also added coverage for closing a session while another coroutine is reading from it. The complete SSH2 suite passes in the PHP 8.4 container. The SCP fixture was moved under `scripts/`, so `git diff --check` is clean.

binaryfire:
@matyhtf When you have a moment, could you please take another look at this PR? I have a separate SSH2 lifecycle fix that depends on the final ownership changes here.

binaryfire:
Follow-up: this branch now also closes the SSH transport descriptor when fatal PHP thread shutdown prevents a bound reader from unwinding. A focused Linux thread regression counts socket descriptors before and after shutdown; it fails on the previous PR head and passes with the fix. I also merged the current master branch.

PR #6143, opened 2026-08-09
Fix ProcessPool listener ownership and partial-start cleanup

binaryfire:
## Summary

ProcessPool publishes listener state before socket creation succeeds. A failed task-worker stream listen can therefore leak the copied address, leave stale state behind, and prevent a clean retry. A second successful listen can also overwrite the active listener.

Build listener state locally and publish it only after socket creation succeeds. Share that setup between the Unix and TCP overloads, and reject attempts to replace an active listener.

Signal-driven log reopening can run before initial worker spawning completes. Skip zero-valued pid slots in `kill_all_workers()` so `kill(0, ...)` cannot target the master process group.

## Testing

- focused ProcessPool core tests

### DIFF
--- a/src/os/process_pool.cc
+++ b/src/os/process_pool.cc
@@ -172,15 +172,18 @@ int ProcessPool::listen(const char *socket_file, int backlog) const {
         swoole_error_log(SW_LOG_WARNING, SW_ERROR_OPERATION_NOT_SUPPORT, "not support, ipc_mode must be SW_IPC_SOCKET");
         return SW_ERR;
     }
-    stream_info_->socket_file = sw_strdup(socket_file);
-    if (stream_info_->socket_file == nullptr) {
+    char *_socket_file = sw_strdup(socket_file);
+    if (_socket_file == nullptr) {
         return SW_ERR;
     }
+    auto _socket = make_server_socket(SW_SOCK_UNIX_STREAM, socket_file, 0, backlog);
+    if (!_socket) {
+        sw_free(_socket_file);
+        return SW_ERR;
+    }
+    stream_info_->socket_file = _socket_file;
     stream_info_->socket_port = 0;
-    stream_info_->socket = make_server_socket(SW_SOCK_UNIX_STREAM, stream_info_->socket_file, 0, backlog);
-    if (!stream_info_->socket) {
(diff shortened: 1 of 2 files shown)


src/os/process_pool.cc
```diff
@@ -172,15 +172,18 @@ int ProcessPool::listen(const char *socket_file, int backlog) const {
...
-    stream_info_->socket = make_server_socket(SW_SOCK_UNIX_STREAM, stream_info_->socket_file, 0, backlog);
-    if (!stream_info_->socket) {
+    auto _socket = make_server_socket(SW_SOCK_UNIX_STREAM, socket_file, 0, backlog);
+    if (!_socket) {
+        sw_free(_socket_file);
         return SW_ERR;
     }
+    stream_info_->socket_file = _socket_file;
+    stream_info_->socket_port = 0;
+    stream_info_->socket = _socket;
```
NathanFreeman: These two parts of the code are very similar and can be combined into a single function to reduce code duplication.

src/os/process_pool.cc
```diff
@@ -1007,6 +1016,9 @@ int ProcessPool::wait() {
     // concurrent kill
     for (i = 0; i < worker_num; i++) {
         worker = &workers[i];
+        if (worker->pid <= 0) {
+            continue;
+        }
```
NathanFreeman: https://github.com/swoole/swoole-src/blob/7430398d324427a932412814afbf746968a086bb/src/core/base.cc#L342-L351  

With this fix in place, there are no longer any cases where a pid is less than zero.

src/os/process_pool.cc
binaryfire: Agreed. I've moved the shared allocation and socket setup into one internal helper and kept the two public overloads as thin delegates.

src/os/process_pool.cc
binaryfire: Agreed for `wait()`. I've removed both guards because it runs only after every worker has been spawned, and `swoole_fork()` doesn't return a negative PID. I've retained the guard in `kill_all_workers()` because signal-driven log reopening can reach it during the initial spawn loop while unspawned worker slots still have a zero PID.

PR #6144, opened 2026-08-09
Fix MessageBus resource cleanup

binaryfire:
## Summary

`MessageBus` owns two kinds of memory: assembled packet payloads, which PHP paths take over as `zend_string` values, and a fixed scratch buffer used for pipe I/O. Both used the single configurable allocator, which is why the destructor could not release the buffer. `ThreadFactory::create_message_bus()` and `ProcessPool::create_message_bus()` allocate it with the standard allocator, and the configured allocator is switched to the Zend one afterwards, so freeing through the configured allocator would release a `malloc()` pointer as a `zend_string`. The buffer was therefore never released for those owners, and the pipe socket wrappers were leaked whenever a reused `ReactorThread` re-initialised its bus.

Give the fixed buffer its own standard allocation path and leave the configurable allocator applying only to assembled packet payloads. The destructor now releases the buffer as well as the pipe wrappers. Reactor threads release both after a normal exit, after a failed `init()`, and before a replacement initialises its reactor. Borrowed pipe descriptors stay open and released slots stay null.

Worker threads also clear pending packets before `php_request_shutdown()`. `Server::call_worker_stop_callback()` already does this on a normal exit, but a bailout skips it and `ReactorThread::init()` does not replace `packet_pool_`, so request-owned strings could otherwise outlive their arena in a bus that is reused.

The fixed IPC buffer is no longer counted as PHP request memory.

## Testing

- focused MessageBus core tests
- PHP 8.4 ZTS build with thread support
- focused thread exit, bailout, replacement, messaging, and reload tests


include/swoole_message_bus.h
```diff
@@ -59,7 +59,7 @@ struct PacketTask {
 
 class MessageBus {
   private:
-    const Allocator *allocator_;
+    const Allocator *packet_allocator_;
```
NathanFreeman: Without changing its name, MessageBus actually has only one memory allocator.

src/protocol/message_bus.cc
```diff
@@ -44,7 +44,8 @@ PacketPtr MessageBus::get_packet() const {
 }
 
 bool MessageBus::alloc_buffer() {
-    void *_ptr = allocator_->malloc(buffer_size_);
+    free_buffer();
```
NathanFreeman: This is unnecessary here. The function is invoked only at process and thread initialization time. Therefore, if it executes again, that implies the process or thread has been terminated and recreated.

src/protocol/message_bus.cc
```diff
@@ -44,7 +44,8 @@ PacketPtr MessageBus::get_packet() const {
...
+    void *_ptr = sw_malloc(buffer_size_);
```
NathanFreeman: Is there any difference here compared to allocating memory through the allocator_?

include/swoole_message_bus.h
binaryfire: Good point. I’ve kept the existing allocator name and setter, and documented that it applies only to assembled packet payloads. The fixed scratch buffer continues using the standard allocator.

src/protocol/message_bus.cc
binaryfire: The worker thread is recreated, but the owning `ReactorThread` and its `MessageBus` are reused. A PHP bailout can bypass `ReactorThread::clean()`, after which the replacement initializes the same bus. Releasing the existing scratch buffer here prevents that allocation from being overwritten and leaked.

src/protocol/message_bus.cc
binaryfire: Yes. PHP paths configure the allocator as `sw_zend_string_allocator()`, whose allocations belong to the current PHP request. That is useful for assembled packet payloads, but the fixed scratch buffer can outlive that request when a `ReactorThread` is reused. `sw_malloc()` gives the scratch buffer the required native lifetime.

matyhtf (changes requested):
Thanks for working on the MessageBus ownership issue. The scratch-buffer lifetime change is directionally useful, but I think the bailout/replacement lifecycle is still incomplete.

1. Request-owned packet buffers can still outlive the PHP request.

`get_packet_buffer()` continues to allocate assembled payloads through `allocator_`, which is configured as `sw_zend_string_allocator()` in PHP server paths. A PHP bailout can jump out of `ReactorThread_onPipeRead()` before `message_bus.pop()` and before `ReactorThread::clean()`/the worker-stop cleanup runs. For example, the command-response path invokes the PHP callback and calls `pop()` only afterwards.

In that case, `php_request_shutdown()` releases the Zend request arena while `packet_pool_` still contains `String` objects pointing into it. A later `clear()` or MessageBus destruction can then release an already-invalid Zend string. Please ensure `packet_pool_` is cleared before `php_request_shutdown()` on both normal return and bailout (for example, through a common thread-request cleanup path).

2. Pipe wrappers are still leaked after bailout/replacement.

The PR correctly notes that a bailout skips `ReactorThread::clean()`. The old reactor is subsequently destroyed, but the reused `ReactorThread::message_bus` still retains its old `pipe_sockets_`. On replacement, `init_pipe_socket()` assigns directly to `pipe_sockets_[pipe_fd]`, overwriting the old pointer. The destructor can only release the newest wrapper, so every bailout/replacement leaks the previous wrappers.

Please clean stale MessageBus state before the replacement calls `swoole_event_init()`, when no reactor is active, or make replacement of an occupied pipe slot explicitly release the old wrapper safely. The current test only covers reinitialization after an explicit `release_pipe_sockets()`, not direct reuse after bailout.

3. Allocation and release must remain symmetric.

If `alloc_buffer()` uses `allocator_->malloc(buffer_size_)`, then `free_buffer()` should use:

```cpp
allocator_->free(buffer_);
```

If the scratch buffer must use the standard allocator to avoid the PHP request lifetime, please represent that allocator explicitly and use the same allocator object for both allocation and release, for example:

```cpp
const Allocator *buffer_allocator_ = sw_std_allocator();

void *_ptr = buffer_allocator_->malloc(buffer_size_);
// ...
buffer_allocator_->free(buffer_);
```

With the current implementation, `sw_malloc()` must be paired with `sw_free()`; using `allocator_->free()` directly with a `sw_malloc()` allocation would itself be an allocator mismatch.

Please also add regression coverage for bailout followed by ReactorThread/MessageBus reuse, including stale packet-pool and pipe-wrapper cleanup.

binaryfire:
@matyhtf I have updated this in 088718508.

- Request-owned packets are now cleared in the common thread request cleanup before `php_request_shutdown()`, including bailout paths.
- Scratch buffers and pipe wrappers are released before a replacement calls `swoole_event_init()`. This runs with no active reactor because `Socket::free()` otherwise defers reclamation. Released pipe slots remain null so lookup behavior is unchanged.
- The scratch buffer now has an explicit standard allocator used for both allocation and release, separate from the configurable packet allocator.

The core tests now cover allocator separation and released pipe lookup behavior. A focused LeakSanitizer comparison also confirms the destructor fix: the merge base leaked two 65,536-byte scratch buffers, while the updated branch did not. The partial-packet bailout window is timing-dependent, so I did not add a PHPT that could pass without exercising it.

NathanFreeman:
Thank you for your work. However, I don't think it's necessary to make changes to `MessageBus` itself. The reasoning is as follows:

- On normal exit, `ReactorThread::clean()` will handle memory cleanup;
- When an abnormal exit triggers a thread restart, `ReactorThread::init()` in `Server::reactor_thread_main_loop` will overwrite the residual invalid memory, so it won't cause any access issues.

That said, the issue of unreleased sockets in `MessageBus` does exist. I suggest keeping the explicit socket release logic and invoking it at an appropriate point.

binaryfire:
@NathanFreeman You're right about the `ReactorThread` buffer after an abnormal exit. `init()` replaces that pointer without reading or freeing it, and the old allocation belongs to the completed PHP request. I've removed that claim from the description.

The reason `MessageBus` itself still has to change is the destructor. `ThreadFactory::create_message_bus()` and `ProcessPool::create_message_bus()` allocate the fixed buffer through the standard allocator. `php_swoole_server_onWorkerStart()`, `php_swoole_server_onUserWorkerStart()`, and `Swoole\Process\Pool::start()` then configure packet allocations to use `sw_zend_string_allocator()`. With a single allocator the destructor cannot safely free the fixed buffer, so it never did. Each task or user worker thread leaks one, including every replacement thread, and each `ProcessPool` leaks one. LeakSanitizer reports two 65,536-byte buffer allocations leaked on the merge base and none on this branch. Giving the fixed buffer its own standard allocation path is what allows the destructor to release it safely.

The explicit pipe socket release follows your suggestion. It runs before a replacement calls `swoole_event_init()`, while no reactor is active, because `Socket::free()` otherwise defers the release to that reactor.

The pre-shutdown packet clear covers the bailout path. `Server::call_worker_stop_callback()` clears the bus on a normal exit, but a bailout skips it and `ReactorThread::init()` does not replace `packet_pool_`.

If you prefer a different way to address these buffer leaks, please let me know.

binaryfire:
@NathanFreeman When you have a moment, could you please take another look at this PR? I have a separate thread-mode MessageBus fix that depends on the final ownership changes here.

NathanFreeman:
Thank you for taking this issue seriously, and for repeatedly revising things based on the review comments. Your analysis of why the destructor cannot release the scratch buffer, and of the pipe wrapper leak after bailout, is accurate, and the LeakSanitizer comparison is also very convincing.

However, after looking into it, we feel that the direction of this PR's fix is not quite right: it splits MessageBus ownership across two sets of allocators, and also adds release logic into the ReactorThread lifecycle, which makes the overall approach considerably more complex than the problem itself requires. We would prefer a simpler approach that handles resource release uniformly at the lifecycle level, rather than making MessageBus bear these semantics itself. The related fix has already been completed in #6237.

PR #6146, opened 2026-08-10
Update embedded library curl feature detection

binaryfire:
The curl constant fix already landed in [swoole/library@6b77c9c](https://github.com/swoole/library/commit/6b77c9c11ca791c77fa13aad72117300071c32ca). `swoole-src` still embeds the generated library from `5cfa605`.

This regenerates `ext-src/php_swoole_library.h` from `6b77c9c`. The embedded library no longer defines `CURLOPT_PREREQFUNCTION` globally when native curl omits it. Guzzle uses that definition for capability detection and otherwise passes an unsupported option while releasing handles.

The PHPT coverage checks both native curl and Swoole's hooked handler. The committed test dependency lock keeps those tests on dependencies compatible with Swoole's PHP 8.2 floor, and the test runner now installs that lock instead of resolving dependencies on every run.

## EDIT:

The previously failing `8.3 - macOS`, `8.4-ubuntu-24.04`, and `8.4-ubuntu-24.04-arm` jobs now pass the tests addressed here. The PHP 8.4 native-detection test skips as intended because native curl already provides the option.


NathanFreeman:
This change needs to be submitted to the https://github.com/swoole/library repository.

binaryfire:
@NathanFreeman The source change has already landed in swoole/library as 6b77c9c. This PR updates the generated embedded copy in swoole-src from 5cfa605 to that revision and adds regression coverage for the embedded runtime. I’ve updated the title and description to make that distinction clearer.

matyhtf:
I will merge first, and then modify some mistakes

PR #6147, opened 2026-08-10
Fix Windows IOCP and interface tests

binaryfire:
The Windows IOCP tests never reached IOCP. `ProcessManager` requires `Swoole\Atomic`, which is unavailable on Windows, and the socket test used the undefined `IPPROTO_TCP` constant.

This replaces the process fixtures with local coroutine socket listeners and bounded cleanup. It also updates the IP and MAC tests to verify the associative interface map returned by the API.

The focused fixture and interface tests pass locally. The Windows workflow will cover the IOCP and Windows adapter paths.

EDIT: The previously failing Unit Test on Windows (windows-2022, ts) job now passes.

### DIFF
--- a/tests/swoole_function/swoole_get_local_ip.phpt
+++ b/tests/swoole_function/swoole_get_local_ip.phpt
@@ -7,10 +7,16 @@ swoole_function: get local ip
 require __DIR__ . '/../include/bootstrap.php';
 
 $ips = swoole_get_local_ip();
-foreach ($ips as $ip) {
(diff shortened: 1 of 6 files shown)


binaryfire:
Superseded. #6189 landed the `IPPROTO_IP` fix and Windows support for `ProcessManager`, and #6193 added Windows atomic wait/wakeup, so the fixtures this PR replaced now work on master and the Windows job exercises them. Nothing here is still needed — closing. Thanks @matyhtf.

PR #6148, opened 2026-08-10
Preserve Atomic wakeups across interrupted waits

binaryfire:
Atomic waits could report failure after a signal interruption, a wakeup race, or a multi-waiter wake. This retries recoverable futex results within the original deadline and resolves the matching timeout race in the polling fallback.

A wakeup that arrives after the kernel has committed a timeout remains pending for the next wait. The futex and polling implementations use different internal mechanisms for `wakeup($n)`, but now expose the same result to callers.

This also removes mixed flag/wait usage from two Table tests that could consume their readiness flag and hang.

The focused Atomic core tests and affected PHPTs pass on both futex and forced polling builds.

### DIFF
--- a/core-tests/src/core/base.cpp
+++ b/core-tests/src/core/base.cpp
@@ -26,6 +26,10 @@
 
 #include "swoole_api.h"
 
+#include <atomic>
+#include <chrono>
+#include <csignal>
+#include <pthread.h>
 #include <sys/resource.h>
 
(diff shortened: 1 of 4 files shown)


NathanFreeman:
This PR will be merged into the `optimize-atomic-wakeups` branch first, and then into the `master` branch after the optimization details are finalized.

PR #6149, opened 2026-08-11
Fix coroutine HTTP client response parsing

binaryfire:
Several coroutine HTTP client behaviors depended on response header casing, field order, or state left by the previous request.

This change:

- parses comma-separated header tokens directly without copying or changing the input
- applies response semantics independently of public header casing and wire order
- scopes connection-close and download state to one exchange
- handles request `Host` keys case-insensitively and emits `Connection: close`
- rejects EOF before a framed response is complete

The tests use local fixtures for request reuse, compression, event streams, WebSocket negotiation, downloads, and truncated responses.


NathanFreeman:
I will merge this PR into the refactory-http-coro-client branch first, and then merge it into the master branch after fine-tuning some details.

PR #6151, opened 2026-08-12
Fix IOCP completion error translation and stale readiness dispatch

binaryfire:
This now builds on #6189, which includes the IOCP cancellation lifetime work originally in this PR.

`GetQueuedCompletionStatus()` reports Win32 system error codes, but completion failures were translated as Winsock errors. A canceled accept therefore exposed raw error `995` instead of `ECANCELED`. Route these failures through the system error mapper and add the network completion codes this path can return.

Non-socket readiness snapshots also retained only descriptor numbers. If a handler replaced a socket with the same descriptor during one dispatch batch, the old readiness result could be delivered to the new socket. Keep each result bound to the socket captured in its snapshot.

The regressions assert the exact cancellation error and descriptor reuse within one readiness batch.


matyhtf (changes requested):
Thanks for addressing the IOCP cancellation lifetime. There is still a reactor-operation lifetime issue during teardown that needs to be fixed before merging.

In `PollOperation::on_complete()`, an orphaned operation is deleted without clearing the corresponding `PollState::operation`:

```cpp
if (reactor && !event->orphaned) {
    reactor->complete(operation, error);
}
delete operation;
```

`Iocp::~Iocp()` marks every pending event as orphaned and drains the completion queue. For an AFD poll, the callback therefore deletes the `PollOperation`, while `states_[fd].operation` continues to point to the freed object.

That stale pointer can still be reached during `Reactor::~Reactor()`: the IOCP destroy callback was registered early and runs before later destroy callbacks. A later callback may call `swoole_event_del()`; for example, the async-thread cleanup callback deletes its Windows socket-pair read socket. `ReactorIocp::del()` then calls `cancel(state.operation)` and dereferences the freed operation.

Please clear the matching state entry for both normal and orphaned completions, while dispatching events only for normal completions. One possible structure is:

```cpp
static void on_complete(IocpEvent *event, DWORD transferred, DWORD error) {
    (void) transferred;
    auto *operation = static_cast<PollOperation *>(event->private_data);

    if (operation->reactor) {
        operation->reactor->complete(operation, error);
    }

    delete operation;
}
```

Then in `complete()`:

```cpp
auto iter = states_.find(fd);
if (iter == states_.end() || iter->second.operation != operation) {
    return;
}

PollState *state = &iter->second;
state->operation = nullptr;

if (operation->event.orphaned) {
    return;
}

// Convert and dispatch the completed events...
```

Please also extend the shutdown regression to cover a later reactor destroy callback removing a socket after IOCP has drained its pending AFD poll. The current test verifies draining, but it does not exercise the stale `PollState::operation` path.

For completeness, the current Windows CI failures are the same three baseline failures seen on `master` (`Swoole\\Atomic` unavailable in the two existing IOCP tests and the existing `local_ip` assertion); they do not appear to be introduced by this PR.

binaryfire:
@matyhtf Fixed in 983126120. Every AFD poll completion now clears the matching reactor state before the operation is released, including during teardown. Orphaned teardown completions clear state without dispatching handlers.

I also extended the Windows shutdown test to cover later async-thread cleanup removing its notification socket after IOCP drains pending work. The updated test passes in the Windows workflow; only the same three baseline failures remain.

binaryfire:
@matyhtf I noticed #6189 incorporated the original IOCP cancellation-lifetime work. I’ve updated this PR against the current master so it now contains only the remaining completion-error translation and stale readiness fixes, with focused regressions. Please take another look when you have a moment.

binaryfire:
Follow-up: This PR has been updated to build on #6189, which contains the earlier Windows cancellation work. It now maps Windows completion failures to the correct portable errors and prevents an event for an old socket from being delivered to a new socket that reused the same descriptor. Tests cover canceled accepts and descriptor reuse, and the descriptor-reuse test now runs in Windows CI.

PR #6153, opened 2026-08-12
Fail tests on missing ProcessManager readiness signals

binaryfire:
`ProcessManager` currently ignores a failed readiness wait. A test then continues after a one-second delay even though its child is not ready, so broken startup ordering can pass, or fail later for the wrong reason.

This makes `ProcessManager` throw when an expected wakeup does not arrive, and terminates and reaps the child it owns when the parent-side barrier fails. Existing fixtures now signal only after their server or required workers are ready. Independent phases use separate `Atomic` values. Two hard-coded listener ports and a missing extension check were also corrected, because those failures would otherwise be reported as a missing wakeup.

This targets `optimize-atomic-wakeups` because the strict check depends on that branch retrying `EAGAIN` and `EINTR` within a single deadline. On `master`, a wakeup that lands between the wait's first check and its `FUTEX_WAIT` call returns `EAGAIN`, which is currently reported as a timeout. It should not move to `master` until that change lands there.

### DIFF
--- a/tests/include/api/http_server.php
+++ b/tests/include/api/http_server.php
@@ -10,9 +10,9 @@ $http->set(array(
     "http_parse_post" => 1,
     "upload_tmp_dir" => "/tmp",
 ));
-$http->on("WorkerStart", function (Swoole\Server $serv) {
+$http->on("WorkerStart", function (Swoole\Server $server, int $workerId) {
     global $pm;
-    if ($pm) {
+    if ($workerId === 0) {
         $pm->wakeup();
     }
 });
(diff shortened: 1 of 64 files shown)


NathanFreeman:
Thank you for your work. Based on the original design intent of Swoole\Atomic as a locking mechanism, its wake-up operation is inherently exclusive. Therefore, the previous modification could not make all awakened processes return true.

binaryfire:
@NathanFreeman You’re right that `Swoole\Atomic` is exclusive, and only the waiter that consumes the token returns true. I’ve updated this PR to work within that model: each `ProcessManager` wait has one matching wakeup, and separate phases use separate `Atomic` values.

The strict check only depends on this branch retrying `EAGAIN` and `EINTR`. It doesn’t require any further `Atomic` changes.

I’ve also removed the unrelated `CurlManager` and file-renaming changes and corrected the affected test handshakes.

PR #6154, opened 2026-08-13
Fix incremental request body and large upload parsing

binaryfire:
`Swoole\Http\Request::parse()` accepts the complete HTTP request header in the first call and request body data in later calls. It already copied every call into a buffer owned by the request, but passed the caller's string to the parser. When a form body arrived over several calls, the stored body pointer could refer outside the current input, so the parser could read outside that input and produce incorrect POST data. This change parses later body data from the request-owned buffer. The complete-header contract is unchanged.

Large upload preprocessing parses multipart data while the upload is still being received. A part header split between reads was either rejected or rebuilt from a pointer into the previous read, so a valid upload could fail or lose fields. A `Content-Disposition` value longer than the fixed 64 KiB scratch buffer was also copied past its end.

This change preserves partial multipart part headers until they are complete, enforces the existing size limits before rebuilding the request, and checks parser initialization and final completion. The bundled multipart parser no longer treats a header value that ends at an input boundary as an error, because mid-stream it cannot tell a truncated request from one that is still arriving. A body that really ends inside a part header is still rejected through the new completion check. Multipart parser initialization now also runs on Windows, where it was compiled out while still reporting success.

It closes unfinished upload state during cleanup. Repeated `Content-Type` headers now replace the parser state derived from the previous value, matching the existing last-value-wins behavior. A failing multipart callback now reports `callback error` instead of an empty reason.

Focused regressions cover a complete request header followed by later body data, multipart part headers split inside both the header name and its value, invalid multipart boundaries, repeated content types, incomplete input, oversized reconstructed headers, and callback failures.


NathanFreeman:
Swoole's asynchronous server waits until the entire request data has been received before it starts parsing. If the reception times out, or if the packet size exceeds the configured limit, the request is discarded directly and will not be processed.

binaryfire:
@NathanFreeman Correct for normal request headers: the async server receives the complete request before the worker parses it.

This PR covers other cases that are parsed in parts:

- `Swoole\Http\Request::parse()` can be called multiple times.
- The coroutine HTTP client can receive response trailers across multiple reads.
- Chunked multipart bodies are passed to the multipart parser one chunk at a time.
- Large multipart uploads are parsed while the body is still being received. This happens in `ListenPort::readable_callback_http()`, which calls `parse_multipart_data()`.

The cleanup and duplicate `Content-Type` fixes are separate from fragmentation. I've updated the PR description to make this clearer.

tests/swoole_http2_server/multipart_fragmented_headers.phpt
```diff
@@ -0,0 +1,60 @@
+--TEST--
+swoole_http2_server: parse multipart headers with shared completion callbacks
```
NathanFreeman: This test can pass without these fixes.

tests/swoole_http_server/multipart_fragmented_headers.phpt
```diff
@@ -0,0 +1,55 @@
+--TEST--
+swoole_http_server: parse fragmented multipart headers
```
NathanFreeman: I understand that this test is designed to verify the combination of Transfer-Encoding: chunked and multipart/form-data. However, I believe this compatibility is unnecessary. For small files, simply submitting them directly via multipart/form-data is sufficient. For large files, a more reliable approach is chunked file upload — where the frontend splits the file into smaller pieces, sends each piece as an independent multipart/form-data request, and the backend reassembles these pieces after receiving all of them.

Transfer-Encoding: chunked is essentially a single, continuous HTTP request. If any minor network fluctuation or interruption occurs during this potentially minutes-long streaming transmission, the entire TCP connection will be terminated, and all the data already transmitted — say, 99% of the file — will be lost entirely, requiring a full retransmission from the very beginning.

tests/swoole_http_client_coro/fragmented_trailer.phpt
```diff
@@ -0,0 +1,55 @@
...
+        $pm->wakeup();
+    });
+    $server->on('receive', function (Swoole\Server $server, int $fd, int $reactorId, string $data) {
+        if (str_starts_with($data, 'GET /oversized ')) {
+            $response = "HTTP/1.1 200 OK\r\nTransfer-Encoding: chunked\r\n\r\n0\r\nX-Trailer: ";
+            $response .= str_repeat('A', 65536) . "\r\n\r\n";
+            $server->send($fd, $response);
+            return;
+        }
+        $server->send($fd, "HTTP/1.1 200 OK\r\nX-Before: intact\r\nTransfer-Encoding: chunked\r\n\r\n4\r\nbody\r\n0\r\nX-Tra");
```
NathanFreeman: Here, 0\r\n indicates the end of the entire message.

tests/swoole_http2_server/multipart_fragmented_headers.phpt
binaryfire: @NathanFreeman You are right. This HTTP/2 test does not depend on the fix, so I have removed it.

tests/swoole_http_server/multipart_fragmented_headers.phpt
binaryfire: @NathanFreeman I understand. Splitting large files into separate requests is more reliable.

This test is not intended to recommend one long chunked upload. It uses behavior that Swoole currently accepts to make the multipart parser receive a header in more than one part.

Do you mean Swoole should not support `Transfer-Encoding: chunked` together with `multipart/form-data`, or only that it should not be recommended for large file uploads?

If this combination is not supported, I will remove these tests. If it is supported, the parser should still handle the request correctly.

tests/swoole_http_client_coro/fragmented_trailer.phpt
binaryfire: @NathanFreeman In HTTP/1.1 chunked encoding, `0\r\n` marks the last data chunk. Trailer fields may follow it, and the message ends after the empty line following the trailers:

```text
0\r\n
X-Trailer: value\r\n
\r\n
```

This test checks a trailer field received in multiple parts. I added `Trailer: X-Trailer` to the initial response headers to make the intent clearer.

ext-src/swoole_http_client_coro.cc
```diff
@@ -352,6 +354,16 @@ class Client {
...
+static bool http_parser_append_header_fragment(Client *http, std::string &buffer, const char *at, size_t length) {
+    size_t header_length = http->tmp_header_field_name.length() + http->tmp_header_field_value.length();
+    if (header_length > SW_HTTP_HEADER_MAX_SIZE || length > SW_HTTP_HEADER_MAX_SIZE - header_length) {
+        http->response_header_too_large = true;
+        return false;
+    }
+    buffer.append(at, length);
+    return true;
+}
+
```
NathanFreeman: This function is unnecessary because when receiving HTTP messages, the size of the request headers is already checked.

https://github.com/swoole/swoole-src/blob/cffd0d0db34d01a5c8d862d71eae87d6166c9ba1/ext-src/swoole_http_client_coro.cc#L1507-L1511

ext-src/swoole_http_client_coro.cc
```diff
@@ -465,17 +477,26 @@ void php_swoole_http_parse_set_cookies(const char *at, size_t length, zval *zcoo
...
+    http->tmp_header_field_name.clear();
+    http->tmp_header_field_value.clear();
     zval *zobject = static_cast<zval *>(http->zobject);
 
-    const char *header_name = http->tmp_header_field_name;
-    size_t header_len = http->tmp_header_field_name_len;
+    const char *header_name = header_name_value.c_str();
+    size_t header_len = header_name_value.length();
+    const char *at = header_value.c_str();
+    size_t length = header_value.length();
```
NathanFreeman: This approach does not have any significant advantages over the original.

ext-src/swoole_http2_server.cc
```diff
@@ -1103,6 +1103,10 @@ static int http2_server_parse_header(
                 }
             } else {
                 if (SW_STRCASEEQ((char *) nv.name, nv.namelen, "content-type")) {
+                    if (ctx->content_type_seen) {
+                        return SW_ERR;
+                    }
+                    ctx->content_type_seen = 1;
```
NathanFreeman: I think this is worth discussing. Currently, Nginx does not forcibly return an error when it encounters duplicate request headers. Swoole's HTTP and HTTP2 servers, on the other hand, handle duplicate request headers by having the latter override the former.

ext-src/swoole_http_client_coro.cc
```diff
@@ -1520,7 +1544,7 @@ bool Client::recv_response(double timeout) {
                     buffer->str + header_crlf_offset, buffer->length - header_crlf_offset, ZEND_STRL("\r\n\r\n")) < 0) {
                 if (buffer->length == buffer->size) {
                     swoole_error_log(SW_LOG_TRACE, SW_ERROR_HTTP_INVALID_PROTOCOL, "Http header too large");
-                    socket->set_err(SW_ERROR_HTTP_INVALID_PROTOCOL);
+                    socket->set_err(SW_ERROR_HTTP_INVALID_PROTOCOL, "HTTP/1 header block is too large");
```
NathanFreeman: This can be left as is. The code above already explains the error message.

ext-src/swoole_http_client_coro.cc
```diff
@@ -1562,7 +1586,11 @@ bool Client::recv_response(double timeout) {
                 break;
             }
         } else {
-            socket->set_err(SW_ERROR_HTTP_INVALID_PROTOCOL);
+            if (response_header_too_large) {
+                socket->set_err(SW_ERROR_HTTP_INVALID_PROTOCOL, "HTTP/1 header block is too large");
+            } else {
+                socket->set_err(SW_ERROR_HTTP_INVALID_PROTOCOL);
+            }
```
NathanFreeman: This check is unnecessary. If we've reached this point, the request header size must already be within the limit.

ext-src/swoole_http_request.cc
```diff
@@ -45,15 +45,26 @@ static int http_request_on_url(llhttp_t *parser, const char *at, size_t length);
...
 
+static bool http_request_append_header_fragment(HttpContext *ctx, std::string &buffer, const char *at, size_t length) {
+    size_t header_length = ctx->current_header_name.length() + ctx->current_header_value.length();
+    if (header_length > SW_HTTP_HEADER_MAX_SIZE || length > SW_HTTP_HEADER_MAX_SIZE - header_length) {
+        return false;
+    }
+    buffer.append(at, length);
+    return true;
+}
+
```
NathanFreeman: https://github.com/swoole/swoole-src/blob/cffd0d0db34d01a5c8d862d71eae87d6166c9ba1/src/server/port.cc#L513-L521

The HTTP request header size has already been checked here.

ext-src/swoole_http_request.cc
```diff
@@ -165,11 +180,16 @@ bool HttpContext::parse_multipart_data(const char *at, size_t length) const {
...
+        if (l_error == 0) {
+            swoole_error_log(
+                SW_LOG_NOTICE, SW_ERROR_SERVER_INVALID_REQUEST, "parse multipart body failed, callback aborted");
+        } else {
+            swoole_error_log(SW_LOG_NOTICE,
+                             SW_ERROR_SERVER_INVALID_REQUEST,
+                             "parse multipart body failed, reason: %.*s",
+                             l_error,
+                             sw_tg_buffer()->str);
+        }
```
NathanFreeman: There's no need to handle the success case separately here.

src/wrapper/http.cc
```diff
@@ -79,23 +81,39 @@ static  constexpr multipart_parser_settings mt_parser_settings = {
...
     llhttp_t parser;
-    multipart_parser *mt_parser;
+    multipart_parser *mt_parser = nullptr;
 
     std::string current_header_name;
+    std::string current_header_value;
     std::string current_form_data_name;
     bool current_part_is_file = false;
-    String *form_data_buffer;
+    std::string form_data_buffer;
```
NathanFreeman: This is Swoole::String, not std::string.

src/wrapper/http.cc
```diff
@@ -79,23 +81,39 @@ static  constexpr multipart_parser_settings mt_parser_settings = {
...
 
+    ~ContextImpl() {
+        if (mt_parser) {
+            if (mt_parser->fp) {
+                fclose(mt_parser->fp);
+            }
+            multipart_parser_free(mt_parser);
+        }
+    }
+
```
NathanFreeman, line 111: https://github.com/swoole/swoole-src/blob/cffd0d0db34d01a5c8d862d71eae87d6166c9ba1/src/protocol/http.cc#L812-L827
This piece of code is responsible for cleaning up mt_parser.

NathanFreeman:
It is recommended not to support the combined use of `Transfer-Encoding: chunked` and `multipart/form-data`, as well as the trailer fields feature (Trailers). The main reasons are that mainstream browsers have limited compatibility with these protocol features, and in mature industry practice, they are typically handled by reverse proxy layers such as Nginx, so there is no need for the application layer to implement them redundantly.

binaryfire:
@NathanFreeman Agreed. I’ve removed response-trailer support and the chunked multipart tests. The PR is now limited to fragmented `Request::parse()` input, large-upload preprocessing, and multipart cleanup. I’ve updated the description to match.

ext-src/swoole_http_client_coro.cc
binaryfire: @NathanFreeman Agreed. I’ve removed this helper and reverted the coroutine HTTP client changes.

ext-src/swoole_http_client_coro.cc
binaryfire: @NathanFreeman Agreed. I’ve restored the original client parser handling here.

ext-src/swoole_http2_server.cc
binaryfire: @NathanFreeman Agreed. I’ve removed the rejection. The later `Content-Type` now also replaces the body parser state, so multipart and URL-encoded parsing follow the same last-wins behavior as the header array.

ext-src/swoole_http_client_coro.cc
binaryfire: @NathanFreeman Agreed. I’ve restored the original error handling here.

ext-src/swoole_http_client_coro.cc
binaryfire: @NathanFreeman Agreed. I’ve removed this check and reverted the client path.

ext-src/swoole_http_request.cc
binaryfire: @NathanFreeman Agreed on the additional size check; I’ve removed it. I kept owned header strings because `Swoole\Http\Request::parse()` can receive separate caller-owned strings across calls, and that path doesn’t go through `ListenPort`.

src/wrapper/http.cc
binaryfire: @NathanFreeman Correct. I’ve restored `Swoole::String`. `ContextImpl` owns it and releases it with the multipart parser.

src/wrapper/http.cc
binaryfire: @NathanFreeman The linked cleanup handles `Request::form_data_->multipart_parser_`. The wrapper uses a separate `ContextImpl::mt_parser`, so that cleanup doesn’t reach it. I’ve kept the wrapper cleanup and made it close any open upload stream before freeing the parser.

ext-src/swoole_http_request.cc
binaryfire: @NathanFreeman Agreed. I’ve restored the original error handling here.

ext-src/swoole_http_request.cc
```diff
@@ -376,9 +393,21 @@ bool swoole_http_token_list_contains_value(const char *at, size_t length, const
...
+    std::string header_value = std::move(ctx->current_header_value);
+    ctx->current_header_name.clear();
+    ctx->current_header_value.clear();
     zval *zheader = ctx->request.zheader;
-    const char *header_name = ctx->current_header_name;
-    size_t header_len = ctx->current_header_name_len;
+    const char *header_name = header_name_value.c_str();
+    size_t header_len = header_name_value.length();
+    const char *at = header_value.c_str();
+    size_t length = header_value.length();
```
NathanFreeman: I don't see this as an improvement over the original code. We're still converting std::string to const char * at the end of the day, and the at parameter received by both http_request_on_header_field and http_request_on_header_value is already a complete segment — there's no fragmentation issue there.

ext-src/php_swoole_http.h
```diff
@@ -186,13 +186,12 @@ struct Context {
...
-    const char *current_header_name;
-    size_t current_header_name_len;
+    std::string current_header_name;
+    std::string current_header_value;
     char *current_form_data_name;
     size_t current_form_data_name_len;
     zval *current_multipart_header;
-    const char *tmp_content_type;
-    size_t tmp_content_type_len;
+    std::string tmp_content_type;
```
NathanFreeman: Why was this changed to std::string here?

ext-src/swoole_http_request.cc
```diff
@@ -325,6 +330,18 @@ bool HttpContext::init_multipart_parser(const char *boundary_str, int boundary_l
...
+        if (mt_parser->fp) {
+            fclose(mt_parser->fp);
+        }
+        multipart_parser_free(mt_parser);
+        mt_parser = nullptr;
+    }
+    delete form_data_buffer;
+    form_data_buffer = nullptr;
+}
+
```
NathanFreeman, line 344: Why is this function needed?

ext-src/swoole_http_request.cc
```diff
@@ -497,22 +530,36 @@ static int http_request_on_headers_complete(llhttp_t *parser) {
...
     size_t value_len;
     int ret = 0;
 
     auto *ctx = static_cast<HttpContext *>(p->data);
+    std::string header_name_value = std::move(ctx->current_header_name);
+    std::string header_value = std::move(ctx->current_header_value);
+    ctx->current_header_name.clear();
+    ctx->current_header_value.clear();
+    const char *at = header_value.c_str();
+    size_t length = header_value.length();
```
NathanFreeman, line 549: This also has the same issue as http_request_on_header_value_complete.

src/protocol/http.cc
```diff
@@ -149,27 +149,59 @@ bool HttpProxy::handshake(const String *recv_buffer) {
...
+    auto *form_data = request->form_data_;
+    size_t header_length = form_data->current_header_name.length() + form_data->current_header_value.length();
+    if (header_length > SW_HTTP_HEADER_MAX_SIZE || length > SW_HTTP_HEADER_MAX_SIZE - header_length) {
+        swoole_error_log(SW_LOG_NOTICE, SW_ERROR_SERVER_INVALID_REQUEST, "multipart header is too large");
+        return false;
+    }
+    buffer.append(at, length);
+    return true;
+}
+
```
NathanFreeman: The method that calls this should be swoole::ListenPort::readable_callback_http, which guarantees that the complete HTTP data will be received, and there will be no fragmentation issues.

NathanFreeman:
I've reviewed this PR, and the main point of contention lies in how HTTP messages are received and parsed. The PR author's approach is to parse the data incrementally using llhttp and the multipart parser as it is being received. However, the actual behavior of the underlying layer is as follows:

    Asynchronous server: ensures that a complete HTTP message is received before triggering parsing;

    Coroutine server: ensures that the headers are fully received, while the remaining data is parsed incrementally as it arrives. Even if a header key is incomplete, the parser will pause and wait for more data to arrive before triggering the corresponding callback.

If the underlying layer does not actually work this way, then the PR author needs to provide a reproducible code example to demonstrate that the current logic does have an issue.

NathanFreeman:
My understanding of the C++ HTTP wrapper module is limited to its use in kernel testing, and I don't believe it's involved in production environments. However, to be on the safe side, I'd like to ask @matyhtf to confirm whether there are any other hidden call paths or indirect usages we should be aware of.

NathanFreeman:
I don't have any objections to the fix for init_multipart_parser.

NathanFreeman:
The fix for Swoole\Http\Request::parse() works by copying and concatenating the incoming data into an internal buffer, completely isolating the parser from external data. This avoids the risk of crashes caused by the external data's lifecycle ending.

```c++
static PHP_METHOD(swoole_http_request, parse) {
    HttpContext *ctx = http_request_get_and_check_context(ZEND_THIS);
    if (UNEXPECTED(!ctx) || ctx->completed) {
        RETURN_FALSE;
    }

    char *str;
    size_t l_str;

    ZEND_PARSE_PARAMETERS_START(1, 1)
    Z_PARAM_STRING(str, l_str)
    ZEND_PARSE_PARAMETERS_END_EX(RETURN_FALSE);

    if (Z_TYPE(ctx->request.zdata) != IS_STRING) {
        ZVAL_STRINGL(&ctx->request.zdata, str, l_str);
    } else {
        size_t len = Z_STRLEN(ctx->request.zdata) + l_str;
        zend_string *new_str = zend_string_alloc(len + 1, false);
        memcpy(new_str->val, Z_STRVAL(ctx->request.zdata), Z_STRLEN(ctx->request.zdata));
        memcpy(new_str->val + Z_STRLEN(ctx->request.zdata), str, l_str);
        new_str->val[len] = 0;
        new_str->len = len;
        zval_dtor(&ctx->request.zdata);
        ZVAL_STR(&ctx->request.zdata, new_str);
    }

    RETURN_LONG(ctx->parse(str, l_str));
}
```

binaryfire:
@NathanFreeman You are right that the normal async and coroutine server paths do not expose this header-fragmentation case.

There are two incremental paths: the public `Swoole\Http\Request::parse()` API and the large-upload preprocessor. Both can call the parser multiple times for one request, so a callback can receive only the part available in the current call.

The corrected PHP test sends one complete form part, then splits the next part's header across two writes. I ran it against the PR base. There it fails because the server answers `400 Bad Request`. With this change it answers `200 OK` and returns the field value and the 131072-byte file.

The C++ test has a different purpose. It checks that the large-upload preprocessor rejects an accumulated part header after it crosses the size limit and does not let the request reach the worker. It passes on both versions because the old parser also rejects the fragmented value, so I am not using it as proof of this fix. Its worker-response check makes sure the test cannot silently stop entering the preprocessor path.

If a client stops sending before the multipart header is complete, the request is not dispatched and no response is sent. The connection remains open until the client closes it or the configured idle timeout expires. If the HTTP message actually completes with an unfinished part header, it is rejected.

While checking your comments, I also found the same issue in incremental URL and form-body parsing. Those are now fixed, and the tests parse the same request at every split point and one byte at a time.

For `Request::parse()`, the old code copied the input into `request.zdata`, but passed the original `str` buffer to the parser. Also, `zdata` is replaced on every call. A callback pointer saved during one call could therefore become invalid during the next call. Data that must survive between calls is now stored in owned strings.

I kept the `init_multipart_parser()` check in the large-upload preprocessor. An invalid multipart boundary can make initialization fail, but this caller previously ignored the result and continued with a null parser. The extension and HTTP/2 callers already check the result.

You are also right that the C++ wrapper is currently used only by the core tests. I removed its wrapper-only behavior changes. The remaining changes are required by the shared parser interface and `parse_cookie` return value, and to initialize and release its own parser state safely. The `parse_cookie` length check remains because the production large-upload preprocessor also calls it.

ext-src/swoole_http_request.cc
binaryfire: @NathanFreeman `at` contains the data available for the current callback. It is not guaranteed to contain the complete header value.

When an `llhttp_execute()` input buffer ends, llhttp delivers the current part of the value. A later call continues the same value. The completion callback is the point where the whole value is known.

Previously, each callback replaced the saved pointer and length. The new code accumulates the parts until completion.

ext-src/php_swoole_http.h
binaryfire: @NathanFreeman The part's `Content-Type` can be parsed in one call, while its `Content-Disposition` is parsed in a later call. The content type is read at that later point, so it cannot remain a pointer into the earlier input buffer.

The regression test splits the input between those two headers.

ext-src/swoole_http_request.cc
binaryfire: @NathanFreeman `multipart_parser_free()` only frees the parser allocation. It does not close the open upload file.

This helper closes the file and releases the parser and its buffer during request cleanup, message completion, and parser replacement.

ext-src/swoole_http_request.cc
binaryfire: @NathanFreeman The multipart parser already allows a field name to arrive in multiple pieces. This change handles field values the same way.

I removed the incomplete-value error because the end of one parser call does not mean the header has ended. When the HTTP message actually completes, the final parser state is checked and an unfinished header is still rejected.

The core tests cover split headers, successful completion, and a message that ends with an incomplete header.

NathanFreeman:
I have already understood the details of this PR. The modification to Swoole\Http\Request::parse() is actually unnecessary.

The reason is that both the HTTP asynchronous server and the HTTP coroutine server have already handled the issue of byte-by-byte reception at the underlying level, so developers do not need to intervene additionally.

As for the use case of Request::parse(), the only scenario I can think of is when manually receiving HTTP messages in a TCP server, where this method might be needed. However, this approach requires the user to ensure that the complete HTTP message has been received, and then pass it to Request::parse() for parsing. In other words, this method is only applicable to this specific scenario where developers need to control data integrity themselves.

```php
<?php
// server
use Swoole\Http\Server;
use Swoole\Http\Request;
use Swoole\Http\Response;

$http = new Server('127.0.0.1', 9501);

$http->on('start', function (Server $server) {
    echo "Swoole http server is started at http://127.0.0.1:9501\n";
});

$http->on('request', function (Request $request, Response $response) {
    var_dump($request);
    $response->header('Content-Type', 'text/plain');
    $response->end('Hello World');
});

$http->start();
```


```php
<?php
// client
use function Swoole\Coroutine\run;
use Swoole\Coroutine\Client;


$fragmented = "POST /upload?first=one&second=two HTTP/1.1\r\n";
$fragmented .= "Host: 127.0.0.1\r\n";
$fragmented .= "Content-Type: application/x-www-form-urlencoded\r\n";
$fragmented .= "Authorization: Bearer fragmented\r\n";
$fragmented .= "X-Custom-Header: custom value\r\n";
$fragmented .= "Upgrade-Insecure-Requests: \r\n";
$fragmented .= "User-Agent: fragmented-agent\r\n";
$fragmented .= "Cookie: first=one; second=two\r\n";
$fragmented .= "Content-Length: 0\r\n\r\n";

run(function() use ($fragmented){
    $client = new Client(SWOOLE_SOCK_TCP);
    if (!$client->connect('127.0.0.1', 9501, 0.5)) {
        exit("error code: {$client->errCode}\n");
    }
    foreach (str_split($fragmented) as $byte) {
            $client->send($byte);
            usleep(100);
    }
    $response = $client->recv();
    echo "response: {$response}\n";
    $client->close();
});
```

binaryfire:
@NathanFreeman I agree that the built-in asynchronous and coroutine HTTP servers handle their own receive buffering. The public `Request::parse()` API seems to have a separate use, though, and it predates this PR.

The coroutine server test [`create_response.phpt`](https://github.com/swoole/swoole-src/blob/c43e6fde0/tests/swoole_http_server_coro/create_response.phpt#L26-L32) creates one `Request` and feeds it each `$conn->recv()` chunk in a loop until the request is complete. [`create_request.phpt`](https://github.com/swoole/swoole-src/blob/c43e6fde0/tests/swoole_http_server/create_request.phpt#L29-L36) splits one request across several `parse()` calls in the same way.

This change does not add that usage. It makes it safe at every input boundary, because llhttp can report a URL, header, or body as more than one piece when the input ends inside it.

If `parse()` should require a complete message, I will remove that part from this PR. Before changing the existing examples, I would need to understand the supported replacement: a raw `Coroutine\Server` handler receives socket chunks and has no HTTP framing option, so it cannot know that a request is complete without parsing HTTP itself. Should that example use `Swoole\Coroutine\Http\Server` instead?

If incremental use should remain supported, this change makes it correct. Which behavior should this API support?

The large-upload preprocessing fix is separate from this question. It uses a normal `Swoole\Http\Server` with no `Request::parse()` involved, and `upload_fragmented_multipart_header.phpt` returns 400 on the PR base and 200 with this change.

NathanFreeman:
The large upload preprocessing is a bug, and I am still carefully studying the details of this part.

NathanFreeman:
`Request::parse()` should keep its existing behavior. When using a TCP server to simulate an HTTP server, it is the user's responsibility to ensure the integrity of the HTTP message. The reason `Request::parse()` supports incremental parsing is that when the user receives a complete HTTP request header, they can first call this function to parse out the `Content-Length`, determine how much more data needs to be received, and then pass the received request body data to `Request::parse()` for further parsing.

Parsing byte by byte is a lazy approach that results in multiple memory allocations and copies, which is not the intended design goal of this API.

The `create_request.phpt` test case itself ensures that the received data is a complete HTTP request, so this incremental receiving step is omitted.

binaryfire:
@NathanFreeman I updated the PR to follow the contract you described.

`Request::parse()` now expects the complete HTTP request header in the first call. I removed the URL and header accumulation, so the request line and headers now use the original input directly, without allocating extra strings. I also removed the byte-by-byte tests.

The opening case in `create_request.phpt` split the request at a random offset between 100 and 600 bytes, which could split inside a header. It now passes the complete request header in one call.

Request body data may still be passed in later calls. The parser reads that data from the buffer owned by the request, so URL-encoded POST data stays valid across calls. The regression sends one complete header first, then the body in later calls.

The multipart changes remain for this body-after-header use and for large upload preprocessing. The upload regression sends one complete form part, then splits the next part header across two reads, once inside the header name and once inside its value. Both fail on the base branch; with this change the server returns the parsed field and the uploaded file.

I also updated the PR title and description to match the final scope.

binaryfire:
Follow-up: I’ve added the missing error reason for multipart callback failures. These failures already rejected the request, but the shared parser left its error reason empty, so the HTTP paths logged a blank reason. The parser now reports `callback error`, with focused coverage for both callback paths; request status and parsing behavior are unchanged.

PR #6155, opened 2026-08-13
Fix segfault when calling Process::signal() in manager process with thread-context

crazywhalecc:
## Description

When Swoole is built with `--enable-thread-context` (`SW_USE_THREAD_CONTEXT`), calling `Swoole\Process::signal()` inside `onManagerStart` crashes the manager process with SIGSEGV.

Reproducer:

```php
<?php
$server = new Swoole\Server('127.0.0.1', 9501, SWOOLE_PROCESS);
$server->set(['worker_num' => 1]);
$server->on('ManagerStart', function () {
    Swoole\Process::signal(SIGTERM, function (int $signo) {});
});
$server->on('Receive', function () {});
$server->start();
```

Result:

```
WARNING Server::signal_handler_child_exit(): Fatal Error: manager process exit. status=0, signal=[Segmentation fault: 11]
```

Backtrace:

```
#0  swoole::Reactor::defer (this=0x0) at src/reactor/base.cc:380
#1  swoole_event_defer (...) at src/wrapper/event.cc:89
#2  zim_swoole_process_signal (...) at ext-src/swoole_process.cc:520
#9  php_swoole_server_onManagerStart (...) at ext-src/swoole_server.cc:1454
```

## Root cause

The manager process is a sync process (`Server::is_sync_process()`), but has no event loop: `SwooleTG.reactor` is `nullptr`. With `SW_USE_THREAD_CONTEXT`, the signal registration in the sync-process branch is deferred via `swoole_event_defer()`, which dereferences the null reactor. Deferring is also semantically wrong here: with no reactor loop, the callback would never run.

Related pr: https://github.com/crazywhalecc/static-php-cli/pull/1229

## Fix

Only defer when a reactor exists; otherwise call `swoole_signal_set()` directly (same behavior as non-thread-context builds).

### DIFF
--- a/ext-src/swoole_process.cc
+++ b/ext-src/swoole_process.cc
@@ -495,14 +495,19 @@ static PHP_METHOD(swoole_process, signal) {
     if (zcallback == nullptr) {
         fci_cache = signal_fci_caches[signo];
         if (fci_cache) {
-#ifdef SW_USE_THREAD_CONTEXT
-            swoole_event_defer([signo](void *) { swoole_signal_set(signo, nullptr); }, nullptr);
-#else
-            swoole_signal_set(signo, nullptr);
-#endif
             signal_fci_caches[signo] = nullptr;
-            swoole_event_defer(sw_callable_free, fci_cache);
             SwooleG.signal_listener_num--;
+            if (swoole_event_is_available()) {
+#ifdef SW_USE_THREAD_CONTEXT
+                swoole_event_defer([signo](void *) { swoole_signal_set(signo, nullptr); }, nullptr);
+#else
+                swoole_signal_set(signo, nullptr);
+#endif
+                swoole_event_defer(sw_callable_free, fci_cache);
+            } else {
+                swoole_signal_set(signo, nullptr);
+                sw_callable_free(fci_cache);
+            }
             RETURN_TRUE;
         } else {
             php_swoole_error(E_WARNING, "unable to find the callback of signal [" ZEND_LONG_FMT "]", signo);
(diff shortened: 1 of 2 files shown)


NathanFreeman (approved):

NathanFreeman:
If possible, could you provide a unit test? Thank You

matyhtf (changes requested):
Thanks for fixing the registration path. There is still a symmetric null-reactor crash in the unregister path, so I don't think this is ready to merge yet.

When `Process::signal($signo, null)` is called in `onManagerStart`, the code at the top of `Process::signal()` still calls `swoole_event_defer()` twice:

```cpp
#ifdef SW_USE_THREAD_CONTEXT
swoole_event_defer([signo](void *) { swoole_signal_set(signo, nullptr); }, nullptr);
#else
swoole_signal_set(signo, nullptr);
#endif

signal_fci_caches[signo] = nullptr;
swoole_event_defer(sw_callable_free, fci_cache);
```

The manager process has no reactor, so both the deferred signal reset and the deferred callable cleanup can dereference a null reactor.

Please handle registration, replacement, and unregistration consistently. When a usable reactor exists, callable destruction should remain deferred so that a callback cannot free its own active `fci_cache`. When there is no reactor, the signal handler reset and callable cleanup need a safe synchronous path. Please also prefer `swoole_event_is_available()` over checking `SwooleTG.reactor` directly.

The regression test should cover the missing lifecycle rather than only duplicating the existing registration test, for example:

```php
$serv->on('ManagerStart', function () use ($pm) {
    Process::signal(SIGUSR1, static function () {
    });

    Assert::true(Process::signal(SIGUSR1, null));
    $pm->wakeup();
});
```

It should verify that the manager remains alive and that the removed callback is not invoked after the signal is sent.

matyhtf (approved):

PR #6157, opened 2026-08-14
Fix HTTP/2 HEAD response framing: end stream on HEADERS, keep Content-Length

alt-program:
### HTTP/2 HEAD requests 

A response to a HEAD request must not include content (RFC 9110 S 9.3.2), but the HTTP/2 server decides END_STREAM purely from the response body length (`send_header()` sets it only when `body->length == 0`). 

A HEAD request handled by ordinary code that calls `$response->end($body)` gets a HEADERS frame without END_STREAM followed by a DATA frame carrying the body. 

nghttp (libnghttp2 strict mode) rejects that stream with PROTOCOL_ERROR. 

Lenient clients such as curl tolerate the extra DATA frame.

### This change

- detects HEAD in the response path and sets END_STREAM on the HEADERS frame;
- leaves Content-Length as the value the equivalent GET would send and sends no DATA frame, which a no-content response is permitted and should do (a non-zero Content-Length with no DATA frames, RFC 9113 S 8.1.1);

- this matches nginx and Apache httpd, which return the header fields (including Content-Length) but no DATA frame.

Non-HEAD responses are unchanged; the added path only branches on the request method.

### Testing

- built the extension and confirmed the framing with nghttp: HEAD returns a HEADERS frame with `flags=0x05` (END_STREAM), Content-Length intact, and no DATA frame, while GET is unchanged.
- added `tests/swoole_http2_server/head_response.phpt` (local server plus coroutine HTTP/2 client, no external dependencies) asserting a HEAD response has status 200, Content-Length equal to the GET body length, and an empty body; it fails without the fix.

### DIFF
--- a/ext-src/php_swoole_http.h
+++ b/ext-src/php_swoole_http.h
@@ -328,7 +328,7 @@ class Stream {
     Stream(const Session *client, uint32_t _id);
     ~Stream();
 
-    bool send_header(const String *body, bool end_stream) const;
+    bool send_header(const String *body, bool end_stream, bool is_head = false) const;
     bool send_body(const String *body,
                    bool end_stream,
                    const std::shared_ptr<Session> &session,
--- a/ext-src/swoole_http2_server.cc
+++ b/ext-src/swoole_http2_server.cc
@@ -508,7 +508,7 @@ bool swoole_http2_server_goaway(HttpContext *ctx, zend_long error_code, const ch
     return rv;
 }
 
-bool Http2Stream::send_header(const String *body, bool end_stream) const {
+bool Http2Stream::send_header(const String *body, bool end_stream, bool is_head) const {
     char header_buffer[SW_BUFFER_SIZE_STD];
     ssize_t bytes = http2_server_build_header(ctx, (uchar *) header_buffer, body);
     if (bytes < 0) {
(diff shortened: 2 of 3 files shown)


NathanFreeman:
This fix is heading in the wrong direction. The core issue is that Swoole, when handling HEAD requests, sends any attached response body along to the client, which may cause strict HTTP clients to reject the response. The correct approach is to modify the Swoole\Http\Response::end() method: when it detects that the current response is for a HEAD request and a response body has been set, it should emit a warning and actively discard the response body, rather than sending it out.

alt-program:
Sounds reasonable. I'll look at patching something that supports Content Length on head requests.

matyhtf:
Thanks for working on this. The direct `$response->end($body)` case is improved, but the current implementation still has blocking problems and should not be merged yet.

1. `write()` followed by `end()` leaves the HTTP/2 stream without `END_STREAM`.

`swoole_http2_server_write()` sends HEADERS without `END_STREAM`, followed by a DATA frame, without checking whether the request method is HEAD. When `end()` later reaches `http2_server_respond()`, `ctx->send_header_` is already true, so no final HEADERS frame is sent. The new `is_head` branch also skips `send_end_stream_data_frame()`, and the stream is then removed locally.

For example:

```php
$response->write('foo');
$response->end();
```

This both sends forbidden response content for HEAD and fails to close the stream on the wire, so a strict client may wait indefinitely.

2. `sendfile()` bypasses `http2_server_respond()` entirely.

`swoole_http2_server_send_file()` still sends the file in DATA frames for a HEAD request, so the same protocol violation remains for:

```php
$response->sendfile($file);
```

This is why HEAD handling should be enforced at the common Response layer instead of only inside `http2_server_respond()`. Please handle all body-producing APIs consistently:

- `end($body)`: warn when a non-empty body is supplied and suppress transmission, while retaining the original representation length for `Content-Length`;
- `write($data)`: warn and discard the data before any DATA frame is sent;
- `sendfile()`: generate the same metadata as GET, including the file length, but do not send the file body;
- keep HTTP/1 and HTTP/2 behavior consistent.

Please also add regression tests for `write() + end()` and `sendfile()`. The current test only covers the direct `end(BODY)` path, so it cannot detect these cases. Ideally, the test should also verify that the stream is actually closed with `END_STREAM`, not only that the Swoole client exposes an empty `data` value.

PR #6158, opened 2026-08-14
Fix masked compressed WebSocket frame unpacking

binaryfire:
## What changed

- unmask a temporary copy before decompressing masked frames
- declare the existing `false` result for invalid frames
- cover compressed frames with and without masking, including frames that cannot be inflated
- compare the existing input-preservation test against a snapshot taken before unpacking, so it can detect a modified input

Masked compressed frames were passed to zlib before their mask was removed, so decompression failed.

The `false` return already existed. Correcting the return type prevents debug PHP builds from aborting when that path is used.

### DIFF
--- a/ext-src/stubs/php_swoole_websocket.stub.php
+++ b/ext-src/stubs/php_swoole_websocket.stub.php
@@ -5,7 +5,7 @@ namespace Swoole\WebSocket {
 		public function push(int $fd, \Swoole\WebSocket\Frame|string $data, int $opcode = WEBSOCKET_OPCODE_TEXT, int $flags = SWOOLE_WEBSOCKET_FLAG_FIN): bool {}
 		public function isEstablished(int $fd): bool {}
 		public static function pack(\Swoole\WebSocket\Frame|string $data, int $opcode = WEBSOCKET_OPCODE_TEXT, int $flags = SWOOLE_WEBSOCKET_FLAG_FIN): string {}
-        public static function unpack(string $data): \Swoole\WebSocket\Frame {}
+		public static function unpack(string $data): \Swoole\WebSocket\Frame|false {}
 		public function disconnect(int $fd, int $code = SWOOLE_WEBSOCKET_CLOSE_NORMAL, string $reason = ""): bool {}
 		public function ping(int $fd, string $data = ""): bool {}
 	}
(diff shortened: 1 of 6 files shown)


matyhtf:
Thanks for fixing the masked compressed frame path. The ownership and failure handling look correct, but the current implementation introduces an avoidable allocation and full payload copy for every compressed frame.

`FrameObject::uncompress()` accepts a `const char *` and does not mutate the input. Therefore, an unmasked compressed frame can still be decompressed directly from `frame.payload`; only a masked compressed frame needs a temporary mutable copy. The current unconditional `ZVAL_STRINGL()` increases both copying cost and peak memory usage for the common unmasked case.

Could you please limit the temporary copy to the masked compressed path? For example:

```cpp
zval zpayload;

if (frame.compressed()) {
    const char *payload = frame.payload;
    zval zmasked_payload;
    bool masked = frame.header.MASK && frame.payload_length > 0;

    if (masked) {
        ZVAL_STRINGL(&zmasked_payload, frame.payload, frame.payload_length);
        WebSocket::mask(Z_STRVAL(zmasked_payload), frame.payload_length, frame.mask_key);
        payload = Z_STRVAL(zmasked_payload);
    }

    bool result = FrameObject::uncompress(&zpayload, payload, frame.payload_length);

    if (masked) {
        zval_ptr_dtor(&zmasked_payload);
    }

    if (sw_unlikely(!result)) {
        swoole_set_last_error(SW_ERROR_PROTOCOL_ERROR);
        RETURN_FALSE;
    }
} else {
    ZVAL_STRINGL(&zpayload, frame.payload, frame.payload_length);
    if (frame.header.MASK && frame.payload_length > 0) {
        WebSocket::mask(Z_STRVAL(zpayload), frame.payload_length, frame.mask_key);
    }
}
```

The new tests and the `Frame|false` return type correction otherwise look good. The Windows and macOS CI failures appear unrelated to this PR.

binaryfire:
@matyhtf You’re right. I’ve changed `unpack()` so unmasked compressed frames are decompressed directly from `frame.payload`. It now creates a temporary copy only when a compressed frame is masked, then releases that copy immediately after decompression.

I also updated the tests to cover masked and unmasked compressed frames, decompression failures, empty payloads, and input preservation.

PR #6159, opened 2026-08-16
[6.3]Preserve Atomic wakeups across interrupted waits

NathanFreeman:


matyhtf:
Thanks for working on this. I found two correctness regressions in the latest optimization that should be addressed before merging.

First, the current futex success condition breaks `wakeup(n)` when more than one waiter is released:

```cpp
if (ret == 0 && sw_atomic_cmp_set(atomic, 1, 0)) {
    return 0;
}
return -1;
```

`FUTEX_WAKE` may release multiple waiters, but only one of them can consume the shared `1` flag. The other waiters were still released successfully by the kernel and must not return `SW_ERR`. This restores the bug that commit `48941646d` was intended to fix.

Please keep the CAS as a best-effort flag consumption and return success for every waiter whose futex call returned zero:

```cpp
if (ret == 0) {
    // Only one waiter needs to consume the shared flag, but every
    // waiter released by FUTEX_WAKE must report success.
    sw_atomic_cmp_set(atomic, 1, 0);
    return 0;
}
```

The core test currently covers only `wakeup(1)`, so it cannot detect this regression. Please add a test with at least two blocked waiters, call `wakeup(2)`, and verify that both waits return `SW_OK`.

Second, the latest commit reverted the polling fallback changes from `48941646d`. Platforms without `HAVE_FUTEX`, including macOS and Windows, still have the timeout/wakeup race where a concurrent wakeup can arrive while the waiter is committing its timeout, but the waiter unconditionally restores the counter and returns failure:

```cpp
sw_atomic_fetch_add(atomic, 1);
return -1;
```

Please restore the conditional timeout commit so that a concurrent wakeup wins the race:

```cpp
if (sw_atomic_cmp_set(atomic, i, i + 1)) {
    return -1;
}
return 0;
```

The visible CI failures appear mostly unrelated, but these two Atomic semantic issues can be determined directly from the implementation.

NathanFreeman:
In the Swoole documentation, Swoole\Atomic can be used as a simple lock. Therefore, we cannot allow all woken-up processes to return true; otherwise, the locking mechanism would not work.

PR #6160, opened 2026-08-16
[WIP][6.1]Fix atomic without futex

NathanFreeman:
In environments without futex support, Swoole\Atomic::wait() and Swoole\Atomic::wakeup() rely on atomic decrement operations on sw_atomic_t to implement the waiting mechanism. This approach introduces a problem: when a waiting process times out and exits prematurely, the sw_atomic_t counter becomes inaccurate. This may cause other processes to acquire the resource before it has actually been released, thereby breaking the correctness of the mutual exclusion guarantee.


matyhtf:
6.1 Branches only accept bug fixes

PR #6161, opened 2026-08-20
Fix large multipart upload preprocessing

binaryfire:
The large-upload preprocessor writes file parts to temporary files before dispatch. A client-supplied `Swoole-Upload-File` header was accepted, and failures before dispatch could leave generated files behind or continue with incomplete upload state.

Reject client-supplied marker headers. Keep temporary files owned by the reactor until dispatch, hand their ownership to the worker before the dispatch call, and remove them when a request is rejected. Reject malformed boundaries, and report temporary-file creation failures as unavailable. Keep an empty file input as `UPLOAD_ERR_NO_FILE` when another file triggers preprocessing.

Tests cover marker ordering, malformed boundaries, rebuilt-body limits, temporary-file failures, empty file inputs, successful dispatch, worker survival, and cleanup.

### DIFF
--- a/src/protocol/http.cc
+++ b/src/protocol/http.cc
@@ -164,6 +164,14 @@ static int multipart_on_header_value(multipart_parser *p, const char *at, size_t
     auto *request = static_cast<Request *>(p->data);
     auto *form_data = request->form_data_;
 
+    if (SW_STRCASEEQ(form_data->current_header_name, form_data->current_header_name_len, SW_HTTP_UPLOAD_FILE)) {
+        swoole_error_log(SW_LOG_WARNING,
+                         SW_ERROR_SERVER_INVALID_REQUEST,
+                         "Bad Request: the reserved header '%s' may not be sent by clients",
+                         SW_HTTP_UPLOAD_FILE);
+        return SW_ERR;
+    }
+
     form_data->multipart_buffer_->append(form_data->current_header_name, form_data->current_header_name_len);
     form_data->multipart_buffer_->append(SW_STRL(": "));
     form_data->multipart_buffer_->append(at, length);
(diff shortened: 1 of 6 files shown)


src/protocol/http.cc
```diff
@@ -163,13 +162,21 @@ static int multipart_on_header_value(multipart_parser *p, const char *at, size_t
...
     auto *form_data = request->form_data_;
+    const auto &header_name = form_data->current_header_name;
 
-    form_data->multipart_buffer_->append(form_data->current_header_name, form_data->current_header_name_len);
+    // Drop any client-supplied copy of the marker; the preprocessor appends its own after saving a file part.
+    // The rebuilt body is re-parsed by this same tokenizer, and both consumers match the name with the same
+    // exact-length case-insensitive compare, so nothing the client writes can reach them as a second marker.
+    if (SW_STRCASEEQ(header_name.c_str(), header_name.length(), SW_HTTP_UPLOAD_FILE)) {
+        return 0;
+    }
```
NathanFreeman, line 173: Regarding the bug related to "large file uploads," I believe the fix should be scoped strictly to the issue itself. The specific approach is: when detecting that the user has manually set SW_HTTP_UPLOAD_FILE, directly raise a warning and terminate the current parsing process, without the need for large-scale refactoring across the entire upload chain.

This fix is suggested to be submitted against the 6.1 branch.

NathanFreeman:
This PR contains an overly broad range of changes and includes multiple modifications that are not directly related to the core fix. It is recommended to split it into several independent PRs by functional module, ensuring that each PR has a single responsibility and a clear focus.

src/protocol/http.cc
binaryfire: @NathanFreeman I narrowed the PR to the large-upload preprocessor fix and retargeted it to 6.1. It now rejects a client-provided `Swoole-Upload-File` header with a warning and a 400 response. The broader work is no longer part of this PR.

binaryfire:
@NathanFreeman I split out the unrelated changes. This PR now contains only the 6.1 large-upload preprocessor fix and its focused regression test.

binaryfire:
Follow-up: This PR now also cleans up temporary upload files when preprocessing fails, reports temporary-file creation failures, and transfers file ownership before dispatch so files aren’t leaked or removed too early. It now also preserves `UPLOAD_ERR_NO_FILE` for an empty file input when another file triggers preprocessing. Tests cover these paths and successful dispatch.

PR #6162, opened 2026-08-20
Restrict multipart upload markers to preprocessed requests

binaryfire:
The large-upload preprocessor replaces file content with an internal `Swoole-Upload-File` marker before dispatch. Other HTTP parsing paths treated a client-supplied header with the same name as a generated file path.

Carry the preprocessed state with the dispatched request and only interpret the marker in that path. Reject an internal marker that has no matching file metadata and remove its generated file.

This builds on #6161, which rejects client marker headers in the preprocessor before dispatch.

Tests cover `Request::create()`, ordinary and coroutine servers, the C++ wrapper, valid preprocessed uploads, and malformed generated metadata.


binaryfire:
Follow-up: This PR now builds on #6161 and only treats Swoole-generated upload markers as file paths on preprocessed requests. It also rejects a generated marker without matching file metadata, removes its temporary file, and cleans up intermediate PHP values when parsing fails. Tests cover direct request creation, ordinary and coroutine servers, the C++ wrapper, valid preprocessed uploads, and malformed metadata.

PR #6165, opened 2026-08-21
fix(websocket): generate fresh masking keys

binaryfire:
WebSocket clients currently reuse the same four-byte masking key for every frame. RFC 6455 section 5.3 requires a fresh, unpredictable key for each frame.

This generates each key from the operating system random source before the frame buffer is touched, so a failure leaves the output buffer unchanged. The response send path now propagates that failure instead of sending an empty buffer. Masked frames require one operating system random read per frame. The unused `SW_WEBSOCKET_MASK_DATA` constant is removed.

The random-byte helper now also stops when a platform source fails or makes no progress, avoiding an endless retry.

While propagating failures, this exposed an old return-value bug in masked header-only encoding. The branch was introduced in `b2629d117`, but its only caller ignored the return and did not request masking at the time. It now reports success consistently with the unmasked path.

The focused regressions cover payload, empty masked frames, and header-only encoding. They fail against the unmodified 6.1 branch and pass here.


binaryfire:
Windows CI now matches the master baseline.

- [Master run 31988380125](https://github.com/binaryfire/swoole-src/actions/runs/31988380125): 15 tests, 3 failures.
- [This PR run 32452945211](https://github.com/swoole/swoole-src/actions/runs/32452945211): 15 tests, 3 failures.

Both runs fail the same existing tests: `iocp_curl.phpt`, `iocp_socket.phpt`, and `local_ip.phpt`. The Windows build and extension loading steps pass.

NathanFreeman:
If this is a bug fix, it needs to be done on the 6.1 branch.

binaryfire:
@NathanFreeman This also includes a Windows random-byte backend, which is why I targeted `master`. No problem—I’ll remove the Windows-specific changes and update this PR for `6.1`.

matyhtf:
Merge-order note: this PR targets 6.1, which does not include the Windows build, so it can land there once its branch is updated and CI is green. Please do not forward-port it to master until #6246 has landed. Without #6246, the combination with #6232 makes every masked WebSocket encode fail on Windows. #6246 now passes the complete Windows CI suite.

PR #6169, opened 2026-08-21
Fix persistent client pooling decisions

binaryfire:
Persistent `Swoole\Client` connections could be pooled after shutdown. A write-shutdown socket can pass the read-side liveness check even though it can no longer send, while a read-shutdown socket can remain pooled holding its descriptor.

A healthy persistent connection could also be discarded when an unrelated operation left an error in the thread-global last-error state. That value does not describe the client being closed. A client whose own connection attempt fails is freed before reaching the pool. Dead pooled sockets are already rejected by the existing liveness check when retrieved.

This closes shutdown clients instead of pooling them and removes the unrelated global error check.

The regressions cover ordinary reuse, read and write shutdown, an unrelated connection error before close, and rejection of dead pooled sockets.

### DIFF
--- a/ext-src/swoole_client.cc
+++ b/ext-src/swoole_client.cc
@@ -1096,9 +1096,10 @@ static PHP_METHOD(swoole_client, close) {
         php_swoole_error(E_WARNING, "client socket is closed");
         RETURN_FALSE;
     }
-    // Connection error, or short tcp connection.
-    // No keep connection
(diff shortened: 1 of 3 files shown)


binaryfire:
Follow-up: This PR now also prevents an unrelated Swoole error from causing a healthy persistent client to be discarded. The pooling decision uses the client’s own shutdown state, while a pooled socket is still checked before reuse. Tests cover unrelated connection errors, ordinary reuse, shutdown sockets, and dead pooled sockets.

PR #6173, opened 2026-08-22
Fix coroutine client setting failures and cleanup

binaryfire:
`Swoole\Coroutine\Client`, `Http\Client`, and `Http2\Client` store socket settings until a socket exists. Errors while applying them were ignored. Invalid proxy settings could connect to the target instead of through the proxy, and `enableSSL()` could continue after its TLS settings were rejected.

Stop when deferred settings are invalid—before connecting, or before the TLS handshake on an upgrade. A TLS settings failure closes the partially configured socket; normal handshake failures keep their existing behavior. A failed `set()` on an established connection returns `false` without closing it.

`Swoole\Coroutine\Client::$fd` now reports `-1` after `close()` and after rejected setup, instead of keeping a released descriptor. Proxy credential failures also distinguish an empty username from a missing password and name the correct setting.

Tests cover invalid proxy, protocol, and TLS settings, cleanup and retry, established connections, descriptor state, and proxy credential warnings.

### DIFF
--- a/ext-src/swoole_client_coro.cc
+++ b/ext-src/swoole_client_coro.cc
@@ -596,8 +596,10 @@ static PHP_METHOD(swoole_client_coro, enableSSL) {
         RETURN_FALSE;
     }
     zval *zset = sw_zend_read_property_ex(swoole_client_coro_ce, ZEND_THIS, SW_ZSTR_KNOWN(SW_ZEND_STR_SETTING), 0);
-    if (php_swoole_array_length_safe(zset) > 0) {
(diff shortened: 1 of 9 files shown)


binaryfire:
Follow-up: This PR now stops the generic, HTTP/1, and HTTP/2 coroutine clients when stored socket settings are invalid, including during enableSSL(). It also resets Client::$fd after close or rejected setup and reports the correct proxy credential setting. Tests cover rejection, cleanup and retry, established connections, descriptor state, and proxy warnings.

PR #6177, opened 2026-08-22
Fix WebSocket ping failure handling

binaryfire:
Response::ping() ignored the result of WebSocket::encode(). When the encoder rejected an oversized control-frame payload, ping() sent the empty buffer and returned true without sending a frame.

It now checks the encoder result, sets SWOOLE_ERROR_WEBSOCKET_PACK_FAILED, and returns false before sending.

The existing ping test covers both sides of the 125-byte control-frame limit and exercises the encoder guard.

### DIFF
--- a/ext-src/swoole_http_response.cc
+++ b/ext-src/swoole_http_response.cc
@@ -1210,7 +1210,11 @@ static PHP_METHOD(swoole_http_response, ping) {
     } else if (ctx->websocket) {
         String *buffer = ctx->get_write_buffer();
         buffer->clear();
-        WebSocket::encode(buffer, ZSTR_VAL(zdata), ZSTR_LEN(zdata), WebSocket::OPCODE_PING, WebSocket::FLAG_FIN);
+        if (sw_unlikely(!WebSocket::encode(
+                buffer, ZSTR_VAL(zdata), ZSTR_LEN(zdata), WebSocket::OPCODE_PING, WebSocket::FLAG_FIN))) {
+            swoole_set_last_error(SW_ERROR_WEBSOCKET_PACK_FAILED);
+            RETURN_FALSE;
+        }
         RETURN_BOOL(ctx->send(ctx, buffer->str, buffer->length));
     } else {
         php_swoole_fatal_error(E_WARNING, "only supports websocket or http2 client");
--- a/tests/swoole_http_server_coro/ping.phpt
+++ b/tests/swoole_http_server_coro/ping.phpt
@@ -7,8 +7,6 @@ swoole_http_server_coro: test ping function
 require __DIR__ . '/../include/bootstrap.php';
 use Swoole\Coroutine\Http\Client;
 use Swoole\Coroutine\Http\Server;
-use Swoole\Http\Request;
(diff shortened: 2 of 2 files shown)


ext-src/swoole_http_response.cc
```diff
@@ -1208,10 +1208,18 @@ static PHP_METHOD(swoole_http_response, ping) {
...
     } else if (ctx->websocket) {
-        String *buffer = ctx->get_write_buffer();
-        buffer->clear();
-        WebSocket::encode(buffer, ZSTR_VAL(zdata), ZSTR_LEN(zdata), WebSocket::OPCODE_PING, WebSocket::FLAG_FIN);
-        RETURN_BOOL(ctx->send(ctx, buffer->str, buffer->length));
+        String *http_buffer = ctx->get_write_buffer();
+
+        zval zpayload = {};
+        ZVAL_STR(&zpayload, zdata);
+
```
NathanFreeman: Compared to the original method, this adds one extra memory allocation and copy.

ext-src/swoole_http_response.cc
```diff
@@ -1208,10 +1208,18 @@ static PHP_METHOD(swoole_http_response, ping) {
...
+        FrameObject frame(&zpayload, WebSocket::OPCODE_PING, WebSocket::FLAG_FIN);
+
+        if (sw_unlikely(!frame.pack(http_buffer))) {
+            swoole_set_last_error(SW_ERROR_WEBSOCKET_PACK_FAILED);
```
NathanFreeman: Since this frame is itself a control frame, it does not need to go through the generic frame generation method.

ext-src/swoole_http_response.cc
```diff
@@ -1208,10 +1208,18 @@ static PHP_METHOD(swoole_http_response, ping) {
     if (ctx->http2) {
         RETURN_BOOL(swoole_http2_server_ping(ctx));
     } else if (ctx->websocket) {
-        String *buffer = ctx->get_write_buffer();
-        buffer->clear();
-        WebSocket::encode(buffer, ZSTR_VAL(zdata), ZSTR_LEN(zdata), WebSocket::OPCODE_PING, WebSocket::FLAG_FIN);
```
NathanFreeman, line 1213: The WebSocket::encode method itself already has a check for the control frame length.

ext-src/swoole_http_response.cc
binaryfire: Updated. I removed the FrameObject wrapper and kept the direct encoder path.

ext-src/swoole_http_response.cc
binaryfire: Agreed. Response::ping() now uses WebSocket::encode() directly and checks its result.

ext-src/swoole_http_response.cc
binaryfire: Agreed. The existing encoder validation remains the single check. Response::ping() now returns false and sets SWOOLE_ERROR_WEBSOCKET_PACK_FAILED when it fails.

PR #6180, opened 2026-08-27
Fix HTTP deflate responses using the raw deflate format

### LINKED ISSUE #6179: Invalid compressed data sent when client asks for "deflate"
Tested using Swoole 6.2.2, with both php 8.3 and 8.5, zlib 1.3.1:

when sending an http request declaring support for 'deflate' compression, and letting Swoole carry out compression server-side, with the following conf:
```
    "http_compression": true,
    "http_compression_level": 6,
    "compression_min_length": 20
```

the resulting response body can not be decompressed using the php function `gzuncompress` - which I normally use for that purpose.

Testing the issue, I found out that the same resp. body can be succesfully decompressed using
```
stream_filter_append($fp, 'zlib.inflate', STREAM_FILTER_READ);
$out = stream_get_contents($fp);
```

While this does solve my present problem, I think that the bug lies within Swoole's usage of the zlib functions.

It might not have surfaced yet because the http spec uses confusing wording related to `deflate` compression, and different server/client implementations in the past have adopted different, incompatible algorithms, resulting in `deflate` basically never being used and `gzip` being always used in its place.

This comment is possibly an indicator of a common mistake when using zlib to compress content for the http `deflate` spec: https://www.php.net/manual/en/function.gzuncompress.php#79042 - it mentions the CRC in use being CRC32 instead of Adler32. Either that, or Swoole is sending the 'raw deflate' data...

PS: a good page to start getting to grips with the http spec `deflate` issues: https://devtoys.pro/blog/gzip-deflate-zlib-formats

lazerg:
## Description

`HttpContext::compress()` initializes zlib with a window size of `-0xf`, which emits a raw deflate stream with no header or checksum. `Content-Encoding: deflate` means the zlib data format of RFC 1950, so a spec-following client cannot read the body: `gzuncompress()` fails, and Swoole's own coroutine HTTP client only reaches it through its raw-deflate retry after a `Z_DATA_ERROR`. The branch now uses `SW_ZLIB_ENCODING_DEFLATE`, the same constant the client already inflates with, so the server sends the zlib header and Adler-32 checksum the encoding calls for.

HTTP/2 responses go through the same function, so they are covered too.

## Testing

`tests/swoole_http_server/compression_deflate.phpt` asks for `Accept-Encoding: deflate` over a raw socket and decompresses the response with `gzuncompress()`. It fails on master and passes with the change. The rest of `tests/swoole_http_server`, plus the compression tests under `tests/swoole_http_server_coro` and `tests/swoole_http2_server`, still pass.

Fixes #6179

### DIFF
--- a/ext-src/swoole_http_response.cc
+++ b/ext-src/swoole_http_response.cc
@@ -530,9 +530,9 @@ bool HttpContext::compress(const char *data, size_t length) {
     else if (compression_method == HTTP_COMPRESS_GZIP) {
         encoding = 0x1f;
     }
-    // deflate: -0xf
+    // deflate: 0xf
     else if (compression_method == HTTP_COMPRESS_DEFLATE) {
-        encoding = -0xf;
+        encoding = 0xf;
     }
 #endif
 #ifdef SW_HAVE_BROTLI
--- a/ext-src/swoole_http_response.cc
+++ b/ext-src/swoole_http_response.cc
@@ -530,9 +530,8 @@ bool HttpContext::compress(const char *data, size_t length) {
     else if (compression_method == HTTP_COMPRESS_GZIP) {
         encoding = 0x1f;
     }
-    // deflate: 0xf
     else if (compression_method == HTTP_COMPRESS_DEFLATE) {
-        encoding = 0xf;
+        encoding = SW_ZLIB_ENCODING_DEFLATE;
     }
 #endif
 #ifdef SW_HAVE_BROTLI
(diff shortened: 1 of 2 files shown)


lazerg:
These three failures do not come from this change. The Windows and macOS suites are already red on master, and the Core Tests job failed on `server.no_idle_task_worker`, which does not touch HTTP compression.

NathanFreeman (approved):

NathanFreeman:
Thank you for your work. If possible, the bug fix needs to be done in the 6.1 branch.

lazerg:
Backport for the 6.1 branch is up: #6192. Same fix, no drift, and the compression tests pass there too.

PR #6183, opened 2026-08-29
Skip stale reactor readiness events

binaryfire:
A native wait can report read and write readiness for the same socket. If the first handler changes the socket's registered events, epoll, poll, and kqueue could still dispatch a later event that is no longer registered. For coroutine sockets, that could try to resume a waiter that has already been cleared.

Check the socket's current event before each dispatch, using the same predicates used when events are registered. Also ignore coroutine callbacks when no waiter remains.

The regression test changes a socket from read/write to read-only inside the read handler and verifies that the pending write event is skipped. The kqueue change is compiled by macOS CI.

This change is independent of #6182.

### DIFF
--- a/core-tests/src/reactor/base.cpp
+++ b/core-tests/src/reactor/base.cpp
@@ -338,6 +338,45 @@ TEST(reactor, poll) {
     reactor_test_func(&reactor);
 }
 
+static void test_stale_event(Reactor::Type type) {
+    UnixSocket p(false, SOCK_STREAM);
+    ASSERT_TRUE(p.ready());
+
+    Reactor reactor(1024, type);
+    reactor.once = true;
+
+    int write_count = 0;
+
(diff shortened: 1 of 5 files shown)


NathanFreeman:
Thank you for your work. However, I believe the changes in this PR are contrary to Swoole's core design philosophy. Swoole adheres to the "deferred recycling" principle—it does not forcibly terminate any event in the current event loop, but rather defers cancellation until after the current round ends, thereby ensuring consistency and data safety in the asynchronous non-blocking model. Furthermore, Swoole exposes control over read/write events to users exclusively through three methods: `Swoole\Event::add()`, `set()`, and `del()`. The underlying implementations of all three methods employ a deferred operation mechanism rather than taking immediate effect.

binaryfire:
@NathanFreeman Got it. Makes sense - thanks for clarifying. Closing.

PR #6184, opened 2026-08-29
Fix poll dispatch after file descriptor reuse

binaryfire:
The poll reactor resolves ready file descriptors through the current registration table. If a callback replaces another socket with the same descriptor, a later result from that poll batch can be sent to the new socket.

Epoll keeps each result bound to the socket stored in the kernel event. Give poll the same guarantee by storing the socket beside each `pollfd`. Existing `removed` checks still suppress sockets removed by an earlier handler, without copying the registration table or repeating its lookups.

Also repair the dead removal loop in the existing poll capacity test so its deletion assertions run.

Poll is the default reactor on macOS and BSD unless kqueue is enabled.

### DIFF
--- a/core-tests/src/reactor/base.cpp
+++ b/core-tests/src/reactor/base.cpp
@@ -338,6 +338,76 @@ TEST(reactor, poll) {
     reactor_test_func(&reactor);
 }
 
+struct PollSocketState {
+    network::Socket *sockets[2];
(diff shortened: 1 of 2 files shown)


binaryfire:
Follow-up: This PR now also repairs the removal loop in the existing poll capacity test, so its deletion assertions actually run. The existing assertions pass unchanged alongside the new descriptor-reuse regression.

PR #6186, opened 2026-08-31
Fix coroutine HTTP client socket initialization

binaryfire:
## Summary

`Swoole\Coroutine\Http\Client` and `Swoole\Coroutine\Http2\Client` leave their internal `zsocket` uninitialized until `connect()` creates a socket. Their public `close()` methods read that value before checking whether the native socket exists. If the leftover bytes look like an object, Swoole takes and releases a reference to a value that was never a valid `zval`.

Initialize both values to null, matching the state restored after a socket closes and the existing `Swoole\Coroutine\Client` behavior. `Swoole\Coroutine\Http\Client::close()` now sets `SWOOLE_ERROR_CLIENT_NO_CONNECTION` when no socket exists, matching its other connection-dependent methods and `Swoole\Coroutine\Client::close()`; before, it returned false without updating `errCode`.

The HTTP/1 PHPT's error assertion fails on unchanged 6.1. Neither PHPT can expose the uninitialized read because whether it causes visible damage depends on the leftover bytes.

The C++ core tests cannot reach PHP-extension `zval` state, normal PHPT assertions cannot inspect it, and Swoole's Valgrind runner suppresses undefined-value reports.

## Reproduction

From the repository root, build without `--enable-asan`:

```sh
phpize
./configure --enable-swoole-dev --enable-debug-log --enable-sockets --enable-mysqlnd --enable-swoole-curl
make -j$(nproc)
```

Then run these commands from the same directory:

```sh
USE_ZEND_ALLOC=0 valgrind --tool=memcheck --track-origins=yes --error-exitcode=99 \
  php -n -d extension=sockets -d extension=modules/swoole.so \
  -r '$client = new Swoole\Coroutine\Http\Client("127.0.0.1", 9501); $client->close();'

USE_ZEND_ALLOC=0 valgrind --tool=memcheck --track-origins=yes --error-exitcode=99 \
  php -n -d extension=sockets -d extension=modules/swoole.so \
  -r '$client = new Swoole\Coroutine\Http2\Client("127.0.0.1", 9501); $client->close();'
```

On unpatched `6.1`, the HTTP/1 command exits 99 and reports:

```text
Conditional jump or move depends on uninitialised value(s)
   at ... zim_swoole_http_client_coro_close (swoole_http_client_coro.cc:2153)
 Uninitialised value was created by a heap allocation
   at ... operator new(unsigned long)
   by ... zim_swoole_http_client_coro___construct (swoole_http_client_coro.cc:1804)
ERROR SUMMARY: 1 errors from 1 contexts
```

The HTTP/2 command has the same trace through `zim_swoole_http2_client_coro_close` at line 1363 and its `new Client` at line 812. These line numbers are from unpatched `6.1`.

With this patch, both commands exit 0 with:

```text
ERROR SUMMARY: 0 errors from 0 contexts
```

### DIFF
--- a/ext-src/swoole_http2_client_coro.cc
+++ b/ext-src/swoole_http2_client_coro.cc
@@ -100,6 +100,7 @@ class Client {
         open_ssl = _ssl;
         _zobject = *zobj;
         zobject = &_zobject;
+        ZVAL_NULL(&zsocket);
         Http2::init_settings(&local_settings);
         local_window_size = local_settings.init_window_size;
     }
--- a/ext-src/swoole_http_client_coro.cc
+++ b/ext-src/swoole_http_client_coro.cc
@@ -583,6 +583,7 @@ Client::Client(const zval *zobject, const std::string &host, zend_long port, zen
     this->ssl = ssl;
 #endif
     _zobject = *zobject;
+    ZVAL_NULL(&zsocket);
     // TODO: zend_read_property cache here (strong type properties)
 }
 
--- a/ext-src/swoole_http_client_coro.cc
+++ b/ext-src/swoole_http_client_coro.cc
@@ -2151,6 +2151,10 @@ static PHP_METHOD(swoole_http_client_coro, recv) {
 
 static PHP_METHOD(swoole_http_client_coro, close) {
     Client *phc = http_client_coro_get_client(ZEND_THIS);
+    if (!ZVAL_IS_OBJECT(&phc->zsocket)) {
+        php_swoole_socket_set_error_properties(ZEND_THIS, SW_ERROR_CLIENT_NO_CONNECTION);
+        RETURN_FALSE;
+    }
     SW_CLIENT_PRESERVE_SOCKET(&phc->zsocket);
     RETURN_BOOL(phc->close());
 }
--- a/tests/swoole_http2_client_coro/close_before_connect.phpt
+++ b/tests/swoole_http2_client_coro/close_before_connect.phpt
@@ -1,0 +1,13 @@
+--TEST--
+swoole_http2_client_coro: close before connect
+--SKIPIF--
+<?php require __DIR__ . '/../include/skipif.inc'; ?>
+--FILE--
+<?php
+require __DIR__ . '/../include/bootstrap.php';
+
+$client = new Swoole\Coroutine\Http2\Client('127.0.0.1', 9501);
+Assert::false($client->close());
+Assert::same($client->errCode, SOCKET_EBADF);
+?>
+--EXPECT--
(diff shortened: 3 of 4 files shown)


NathanFreeman:
Thank you for your work on this. However, I would like to point out that the tests attached to this PR can still pass without this fix. This is because Swoole\Coroutine\Http\Client and Swoole\Coroutine\Http2\Client adopt a lazy-loading mechanism—the socket is not actually created until a formal connection is established (i.e., when connect() is called). Therefore, the current test coverage is insufficient to validate this fix.

binaryfire:
@NathanFreeman You're right that both PHPTs pass without the patch. They add missing coverage for calling `close()` before `connect()`, but they do not prove the initialization fix by themselves.

The source issue is still present because `close()` reads the internal `zsocket` before checking whether the lazily created native socket exists. If the leftover bytes look like an object, Swoole takes and releases a reference to a value that was never a valid `zval`.

The C++ core tests cannot reach this PHP-extension state, normal PHPT assertions cannot inspect it, and Swoole's Valgrind runner suppresses undefined-value reports. I reproduced it directly from the repository root with a developer build, without `--enable-asan`:

```sh
phpize
./configure --enable-swoole-dev --enable-debug-log --enable-sockets --enable-mysqlnd --enable-swoole-curl
make -j$(nproc)

USE_ZEND_ALLOC=0 valgrind --tool=memcheck --track-origins=yes --error-exitcode=99 \
  php -n -d extension=sockets -d extension=modules/swoole.so \
  -r '$client = new Swoole\Coroutine\Http\Client("127.0.0.1", 9501); $client->close();'

USE_ZEND_ALLOC=0 valgrind --tool=memcheck --track-origins=yes --error-exitcode=99 \
  php -n -d extension=sockets -d extension=modules/swoole.so \
  -r '$client = new Swoole\Coroutine\Http2\Client("127.0.0.1", 9501); $client->close();'
```

On unpatched `6.1`, each command exits 99. The HTTP/1 result includes:

```text
Conditional jump or move depends on uninitialised value(s)
   at ... zim_swoole_http_client_coro_close (swoole_http_client_coro.cc:2153)
 Uninitialised value was created by a heap allocation
   at ... operator new(unsigned long)
   by ... zim_swoole_http_client_coro___construct (swoole_http_client_coro.cc:1804)
ERROR SUMMARY: 1 errors from 1 contexts
```

HTTP/2 has the same trace through its `close()` and `new Client`. With this patch, both commands exit 0 with `ERROR SUMMARY: 0 errors from 0 contexts`.

`Swoole\Coroutine\Client` already initializes its equivalent value to null, and both HTTP clients restore the same null state after a socket closes. I've updated the PR description to distinguish the PHPT path coverage from the Memcheck proof.

I've kept the tests because nothing else covers `close()` before `connect()`. I can remove them if you'd rather not carry tests that only cover the path.

binaryfire:
Follow-up: This PR now also makes the HTTP/1 client set `SWOOLE_ERROR_CLIENT_NO_CONNECTION` when `close()` is called without a socket, matching its other connection-dependent methods. The existing close-before-connect test now proves that error contract and fails on unchanged 6.1.

PR #6192, opened 2026-09-06
Fix HTTP deflate responses using the raw deflate format

### LINKED ISSUE #6179: Invalid compressed data sent when client asks for "deflate"

lazerg:
Backports #6180 to the 6.1 branch. HttpContext::compress() set the deflate window size to -0xf, which emits a raw deflate stream instead of the zlib format that Content-Encoding: deflate calls for. gzuncompress() could not read the body, even though gzinflate() and Swoole's own coroutine client still worked, since the client retries with raw deflate after a Z_DATA_ERROR. The branch now uses SW_ZLIB_ENCODING_DEFLATE, the same constant the client already inflates with.

Same root cause and fix as #6180. The code here matches master, so nothing needed adapting. Fixes #6179.

### DIFF
--- a/ext-src/swoole_http_response.cc
+++ b/ext-src/swoole_http_response.cc
@@ -531,9 +531,8 @@ bool HttpContext::compress(const char *data, size_t length) {
     else if (compression_method == HTTP_COMPRESS_GZIP) {
         encoding = 0x1f;
     }
-    // deflate: -0xf
     else if (compression_method == HTTP_COMPRESS_DEFLATE) {
-        encoding = -0xf;
+        encoding = SW_ZLIB_ENCODING_DEFLATE;
     }
 #endif
 #ifdef SW_HAVE_BROTLI
--- a/tests/swoole_http_server/compression_deflate.phpt
+++ b/tests/swoole_http_server/compression_deflate.phpt
@@ -1,0 +1,51 @@
+--TEST--
+swoole_http_server: deflate response uses the zlib data format
+--SKIPIF--
+<?php
+require __DIR__ . '/../include/skipif.inc';
+skip_if_constant_not_defined('SWOOLE_HAVE_ZLIB');
+?>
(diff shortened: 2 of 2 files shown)


lazerg:
The remaining CI failures predate this change and aren't caused by it. The zts-thread jobs fail on invalid PHP version setup in the workflow. The rest (macOS, ubuntu, Hyperf 8.1) fail on unrelated tests such as http2 headers, curl, and server dispatch, none of which touch compression.

NathanFreeman (approved):

PR #6198, opened 2026-09-07
Fix memory errors in Server cpu_affinity_ignore parsing

binaryfire:
`Server::set()` sized the `cpu_affinity_available` allocation from the raw number of `cpu_affinity_ignore` entries, then filled it from the unique valid CPU IDs. Duplicate, negative, and out-of-range IDs could make those counts disagree and write past the allocation. A non-array value was read as an array and crashed. The stored allocation was also not released when the server was destroyed.

This requires the option itself to be an array, ignores unusable CPU IDs, and counts each valid ID once. When no usable custom set remains, it stores a null pointer and zero count, matching the representation expected by the consumers that existed before #5320. Server destruction now releases the stored allocation.

`Server::set()` still accepts and parses this option, although the parsed result has had no consumers since #5320.

The focused PHPTs fail on unchanged 6.1 and pass with this change. The memory reports require an ASAN build:

```sh
phpize
./configure --enable-asan --enable-openssl --enable-sockets
make -j$(nproc)
```

On unchanged 6.1, ASAN reports a null read for a non-array value and a heap-buffer-overflow for duplicate and invalid CPU IDs. LeakSanitizer reports the stored allocation when a configured server is started and stopped. These reports are gone with this change.

### DIFF
--- a/ext-src/swoole_server.cc
+++ b/ext-src/swoole_server.cc
@@ -2223,31 +2223,36 @@ static PHP_METHOD(swoole_server, set) {
     }
     // cpu affinity set
     if (php_swoole_array_get_value(vht, "cpu_affinity_ignore", ztmp)) {
-        int ignore_num = zend_hash_num_elements(Z_ARRVAL_P(ztmp));
-        if (ignore_num >= SW_CPU_NUM) {
-            php_swoole_fatal_error(E_ERROR, "cpu_affinity_ignore num must be less than cpu num (%d)", SW_CPU_NUM);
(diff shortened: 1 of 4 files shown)


binaryfire:
@NathanFreeman I’ve applied the approach described in your comment on #6204 to this PR. Invalid settings are now ignored, valid unique CPU IDs are preserved, and unusable configurations fall back to normal CPU selection. The safe allocation sizing and cleanup remain to address the out-of-bounds write and leak. I’ve also simplified the tests and updated the PR description. Please take another look when you have a moment.

PR #6199, opened 2026-09-07
Fix coroutine HTTP server request boundaries and 100 Continue

binaryfire:
`Swoole\Coroutine\Http\Server` previously looked only at the last five received bytes to decide whether a chunked request was complete. Those bytes can appear inside the body, while trailers and pipelined requests put other bytes after the real ending. Large chunked requests also stopped when they filled the initial receive buffer, even below `package_max_length`.

Header and body size checks had related boundary errors. Oversized headers were not checked, large `Content-Length` values could wrap the total size, and transfer-encoding lists ending in `chunked` were not recognized correctly.

The coroutine server also waited for declared-length and chunked request bodies without sending `HTTP/1.1 100 Continue`. Clients that wait for that interim response stalled until their fallback timeout.

This change parses chunked input as it arrives and uses the parser-reported completion point as the exact end of each request, preserving following pipelined bytes. It validates header and body sizes before arithmetic or parsing can cross the configured limit, recognizes chunked framing from the parser, and sends `100 Continue` once before waiting for either body framing mode.

Chunked multipart bodies are parsed once after dechunking completes, so part headers may span chunk and network boundaries. `Request::getData()` now contains only the current complete request, matching the ordinary HTTP server.

The tests cover large and fragmented chunked bodies, multipart input, trailers, pipelines, exact limits, overflowing declared lengths, declared-length and chunked expect-continue handshakes, one-shot interim responses, and requests without `Expect`. Each regression fails on the unchanged branch and passes with the fix.

### DIFF
--- a/ext-src/swoole_http_request.cc
+++ b/ext-src/swoole_http_request.cc
@@ -403,10 +403,6 @@ static int http_request_on_header_value(llhttp_t *parser, const char *at, size_t
         ctx->set_compression_method(at, length);
     }
 #endif
-    else if (SW_STRCASEEQ(header_name, header_len, "transfer-encoding") && SW_STR_ISTARTS_WITH(at, length, "chunked")) {
-        ctx->recv_chunked = 1;
-    }
-
 _add_header:
     zval tmp;
     ZVAL_STRINGL(&tmp, (char *) at, length);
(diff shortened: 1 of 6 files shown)


binaryfire:
Follow-up: This PR now determines the exact end of each chunked request instead of looking only at the last bytes received. This prevents early completion, accepts trailers, preserves following pipelined requests, keeps `Request::getData()` limited to the current request, and safely enforces `package_max_length` for headers and bodies. It also sends `HTTP/1.1 100 Continue` once before waiting for either a declared-length or chunked body, so clients do not wait for their fallback timeout. Tests cover fragmented input, multipart bodies, trailers, pipelines, exact limits, overflowing `Content-Length` values, and both expect-continue body modes.

PR #6200, opened 2026-09-07
Handle fragmented HTTP proxy responses

binaryfire:
HTTP proxy responses can be split across reads. The synchronous client reads once, and the async client treats an incomplete response as invalid.

The response parser also accepts status 200 before the headers end. If tunneled data arrives in the same read, the client discards it with the proxy response.

This change distinguishes incomplete and invalid responses, waits for the complete proxy header, and consumes only that header. Data from the tunneled connection remains available to the next receive.

Both async proxy failure tests previously asserted only inside `onError`, so they could pass without executing their error checks. Each now asserts after the event loop that the callback ran.

The proxy framing paths are covered across synchronous, async, and coroutine clients with local fixtures. The two existing async failure tests also pass on the updated branch.

### DIFF
--- a/core-tests/src/network/client.cpp
+++ b/core-tests/src/network/client.cpp
@@ -274,7 +274,7 @@ TEST(client, async_tcp_http_proxy_handshake_fail) {
 
     ac.onError = [&success](Client *ac) {
         DEBUG() << "connect failed, ERROR: " << errno << "\n";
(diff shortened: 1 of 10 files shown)


binaryfire:
Follow-up: I added the missing final assertion to the adjacent async SOCKS5 proxy failure test. Both async proxy failure tests now prove that `onError` ran instead of passing when none of their callback assertions executed. The PR description has been updated.

PR #6204, opened 2026-09-08
Fix CPU affinity IDs

binaryfire:
`Swoole\Process::setAffinity()` validates converted CPU IDs but uses the original zval storage when building the affinity mask. Numeric-string IDs can therefore be omitted, and a mixed list can return success with the wrong mask. Negative IDs are also silently omitted on this branch.

The affinity helpers also treat `swoole_cpu_num()` as the CPU ID range, but that count does not bound IDs. On musl it counts the startup affinity mask, and on master it can be a cgroup CPU quota. A process limited to CPU 1 on Alpine gets an empty `getAffinity()` result, and `setAffinity([1])` fails.

This change converts each entry once and uses that value for validation, diagnostics, and the CPU mask. Input IDs and the output scan are limited by the platform CPU-set size; the kernel still rejects a mask with no usable CPU. `Thread::setAffinity()` also reports a rejected mask instead of returning true.

The existing affinity test uses the process's current CPUs directly. A Linux regression compares the affinity with `Cpus_allowed_list` when it includes an ID outside the `swoole_cpu_num()` range.

### DIFF
--- a/ext-src/swoole_process.cc
+++ b/ext-src/swoole_process.cc
@@ -960,8 +960,7 @@ bool php_swoole_array_to_cpu_set(const zval *array, cpu_set_t *cpu_set) {
         return false;
     }
 
-    const zend_long cpu_num = SW_MIN((zend_long) SW_CPU_NUM, (zend_long) CPU_SETSIZE);
-    if ((zend_long) php_swoole_array_length(array) > cpu_num) {
+    if (php_swoole_array_length(array) > CPU_SETSIZE) {
         php_swoole_fatal_error(E_WARNING, "More than the number of CPU");
         return false;
     }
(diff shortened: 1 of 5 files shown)


NathanFreeman:
Thank you for your contribution. However, the implementation of this PR seems a bit overcomplicated. I would suggest simplifying the approach: it requires checking whether the value is an array type. For invalid values (such as negative numbers or configurations that exceed the current CPU core count), simply ignore them and fall back to defaults.

One practical scenario worth considering: an application originally deployed on a 16-core server is later migrated to an 8-core server a few years down the road. In this downgrade migration scenario, under the current PR logic, the value 16 recorded in the configuration file would cause an error because it exceeds the actual core count, preventing the service from starting. This would create unnecessary operational friction for the DevOps team.

binaryfire:
@NathanFreeman Thanks. I think this comment may have been intended for #6198, which handles the `cpu_affinity_ignore` server setting. This PR fixes `Process::setAffinity()` using the original zval after validating its converted value. Unchanged 6.1 already requires an array and rejects CPU IDs at or above the current CPU count, so this PR doesn’t introduce the migration behavior described here. Could you confirm which PR you meant?

binaryfire:
Follow-up: this PR now also uses the platform CPU-set size for affinity IDs. This fixes `getAffinity()` and `setAffinity()` when the process's allowed CPUs don't start at zero, and it also covers quota-derived CPU counts when the fix is merged forward. The new Linux regression compares the result with the kernel's allowed CPU list, and the existing test no longer filters those IDs out.

PR #6205, opened 2026-09-08
Handle HTTP body boundaries and 100 Continue

binaryfire:
The ordinary HTTP server only recognized the end of a chunked request when the receive buffer ended with `0\r\n\r\n`. Requests with trailers or another pipelined request could therefore remain incomplete. Valid chunk extensions were rejected, bytes after body requests were discarded, and large declared lengths could overflow the size calculation. The server also omitted `100 Continue` before chunked and streaming multipart bodies and repeated it as fixed-length body fragments arrived.\n\nThis change parses chunk framing incrementally through extensions and trailers, preserves bytes for the next request after both chunked and `Content-Length` bodies, and checks request sizes without overflowing or growing past `package_max_length`. It sends `100 Continue` while only the request headers are buffered, before waiting for the body.\n\nThe tests cover split framing, extensions, trailers, body request pipelines, malformed chunks, impossible declared lengths, non-aligned package limits, and `100 Continue` across the supported body forms.

### DIFF
--- a/include/swoole_http.h
+++ b/include/swoole_http.h
@@ -112,6 +112,18 @@ namespace swoole {
 class Server;
 namespace http_server {
 //-----------------------------------------------------------------
+enum ChunkState {
+    CHUNK_STATE_SIZE,
+    CHUNK_STATE_DATA,
+    CHUNK_STATE_TRAILER,
+};
+
+enum ChunkSizeResult {
+    CHUNK_SIZE_OK,
+    CHUNK_SIZE_MALFORMED,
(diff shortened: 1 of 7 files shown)


binaryfire:
Follow-up: This PR now also finds the exact end of chunked requests, including extensions and trailers, and preserves a following request after either a chunked or Content-Length body. It rejects impossible declared sizes without overflow. It now sends `100 Continue` before waiting for chunked and streaming multipart bodies, and avoids repeating it as a fixed-length body arrives in fragments. The tests cover these cases.

PR #6210, opened 2026-09-08
Fix reactor initialization and hook tests

binaryfire:
When an event backend cannot initialize, `swoole_event_init()` returns an error without freeing the new reactor. This leaves both the reactor and its backend allocation unreachable.

Delete the failed reactor before returning. Initialize the kqueue event pointer to null because this cleanup makes its failure destructor reachable for the first time.

The reactor hook test also cleared both global hook pointers so its poll-only assertions could not affect later tests. That made its two hook lists unreachable before `swoole_clean()` could free them. Delete the lists before clearing the pointers. The same fixture registered its create hook after constructing the reactor, so that callback never ran; register both hooks first and assert each runs once.

The regression forces `epoll_create()` to fail with `EMFILE`, verifies the reactor is destroyed, and confirms normal initialization still works afterward. The full reactor test group reports 144 leaked bytes under ASAN before the hook cleanup and no leak afterward.

### DIFF
--- a/core-tests/src/reactor/base.cpp
+++ b/core-tests/src/reactor/base.cpp
@@ -550,6 +550,8 @@ TEST(reactor, hook) {
         1);
 
     ON_SCOPE_EXIT {
+        delete static_cast<std::list<Callback> *>(SwooleG.hooks[SW_GLOBAL_HOOK_ON_REACTOR_CREATE]);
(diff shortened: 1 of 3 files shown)


binaryfire:
Follow-up: this PR now also releases the hook lists owned by the reactor hook test before clearing their global pointers. The previous test cleanup made those lists unreachable; the full reactor test group reported 144 leaked bytes under ASAN. It also registers the lifecycle hooks before constructing the reactor and verifies that both callbacks run. The title and description have been updated.

PR #6211, opened 2026-09-08
Fix test fixture isolation and verification

binaryfire:
This fixes test paths that could kill unrelated processes, hang indefinitely, interfere with parallel tests, or pass without exercising the intended verification.

The PHPT runner no longer sweeps host-wide PHP processes or kills whichever process owns a reused port. On Unix, it replaces the shell with the test process. When a test times out, the runner stops the process tree it started on Unix and Windows, so server children cannot continue after the runner moves on.

`CurlManager` now bounds CLI server startup, replaces the shell with PHP, and stops and reaps that exact child on startup failure or a callback exception. Callback exceptions are rethrown after cleanup.

The HTTP/2 fixtures run `nghttpd` as the `ProcessManager` child instead of killing every `nghttpd` process on the host. The worker replacement tests use unique process names, select the task worker correctly, and allow the replacement process to start before checking it.

The client CA fixture now uses `ssl_client_cert_file`, so it actually requests and verifies the client certificate. Its CA and valid client certificates have been renewed; the CA key and subject key identifier are unchanged, and the existing server and client chains still verify.

Six tests that depended on the unavailable `httpcan.org` service now use repository-owned local endpoints while preserving the behavior each test covers.

Focused checks reproduced the callback leak, unavailable-server hang, host-wide `nghttpd` cleanup, vacuous client certificate test, parallel worker-test failures, a timed-out child process surviving, and the six unavailable endpoint failures on the unchanged branch. The updated paths pass with the fixes. The process-tree timeout probe was run on Unix; the Windows path was checked against the runner and PHP process model.


binaryfire:
CI note: the Core Tests (`ubuntu`) check was cancelled by GitHub. I cannot rerun an upstream workflow without repository admin access. The same check passed on the immediately preceding commit: https://github.com/swoole/swoole-src/actions/runs/34265658456. The only later change is in `tests/run-tests`; both Linux thread jobs, Windows, io_uring, and the build jobs passed on the current commit.

binaryfire:
Follow-up: I expanded this PR with related test fixture fixes. `CurlManager` now has bounded startup and always cleans up its server, the HTTP/2 tests stop only the `nghttpd` process they started, worker-replacement tests no longer collide when run in parallel, and the client CA test now performs the certificate verification it was intended to cover. The PHPT timeout path now stops only the process tree started by the timed-out test, so `ProcessManager` children cannot outlive it. Six tests that depended on `httpcan.org` now use local fixtures, so that unavailable service no longer makes those tests fail. The title and description have been updated to reflect the broader scope.

PR #6213, opened 2026-09-08
Keep uploaded file paths valid through request shutdown

binaryfire:
## Summary

- keep uploaded file paths alive while PHP's uploaded-file table references them
- share the same path string with the request temporary-file list
- cover a preprocessed upload retained until request shutdown

## Why

Swoole stored a pointer to temporary stack or heap storage as each uploaded-file table value. Normal
`is_uploaded_file()` and `move_uploaded_file()` calls only inspect the table keys, so they did not expose the lifetime
error. When an upload request survives until extension shutdown, PHP reads the value while removing the temporary file
and can access storage that has already been released.

This can occur through public `Request::create()` and `parse()` usage, a single-worker base-mode server, or thread
mode. The change stores the path in one `zend_string` shared by the request and the uploaded-file table. The table key
then keeps its value valid until PHP finishes cleanup.

The regression uses the real large-upload preprocessor in single-worker base mode and retains the request through
shutdown. It passes normally because ordinary assertions cannot observe the stale native pointer.

## Reproduction

After building and installing Swoole, run the regression under Valgrind:

```sh
cd tests
PHPT=1 php run-tests -m swoole_http_server/upload_file_request_shutdown.phpt
```

On unchanged 6.1, Valgrind reports `unlink(pathname) points to unaddressable byte(s)` from
`destroy_uploaded_files_hash()`. With this change, the test passes cleanly.

### DIFF
--- a/ext-src/swoole_http_request.cc
+++ b/ext-src/swoole_http_request.cc
@@ -321,9 +321,12 @@ void swoole_http_parse_cookie(zval *zcookies, const char *at, size_t length) {
 static void http_request_add_upload_file(HttpContext *ctx, const char *file, size_t l_file) {
     zval *zfiles = swoole_http_init_and_read_property(
         swoole_http_request_ce, ctx->request.zobject, &ctx->request.ztmpfiles, SW_ZSTR_KNOWN(SW_ZEND_STR_TMPFILES));
-    add_next_index_stringl(zfiles, file, l_file);
-    // support is_upload_file
-    zend_hash_str_add_ptr(SG(rfc1867_uploaded_files), file, l_file, (char *) file);
+    zend_string *path = zend_string_init(file, l_file, 0);
+    add_next_index_str(zfiles, zend_string_copy(path));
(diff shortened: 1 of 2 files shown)


NathanFreeman:
Thank you for your contribution. However, I noticed that the unit test still passes without this patch. I looked at the test code and saw a line RequestHolder::$request = $request; in it. Is this fix specifically for that case?

binaryfire:
@NathanFreeman Yes, retaining the request is deliberate. Without that line, the request destructor removes the uploaded-file entry before extension shutdown reaches it.

A normal run passes before and after because the defect is an invalid native-memory read. It reproduces under Valgrind:

```sh
cd tests
PHPT=1 php run-tests -m swoole_http_server/upload_file_request_shutdown.phpt
```

Unchanged 6.1 reports `unlink(pathname) points to unaddressable byte(s)` from `destroy_uploaded_files_hash()`. With this patch, the test passes cleanly. The same lifetime can also occur with `Request::create()`/`parse()` and other requests retained through shutdown, so it isn't specific to the test holder.

I've updated the PR description with these reproduction details.

PR #6214, opened 2026-09-08
Fix server SSL context initialization

binaryfire:
SNI-only server configuration skipped creation of the primary SSL context, so connections failed before SNI could select a certificate. Invalid default certificates were also reported as successfully initialized, and other creation failures left an incomplete context marked ready.

This always creates the primary context, allowing it to omit a certificate only when valid SNI certificates exist. Invalid or incomplete certificate configuration is rejected. Failed context creation now frees and clears the incomplete context so later attempts start cleanly.

The SNI-only and invalid-certificate regressions fail on unchanged 6.1 and pass with this change. Existing SNI and HTTPS server tests continue to pass.

### DIFF
--- a/core-tests/src/protocol/ssl.cpp
+++ b/core-tests/src/protocol/ssl.cpp
@@ -67,9 +67,16 @@ TEST(ssl, get_error) {
 }
 
 TEST(ssl, password) {
-    SSLContext ctx;
+    SSLContext ctx{};
     ctx.key_file = swoole::test::get_ssl_dir() + "/passwd_key.pem";
     ctx.passphrase = "swoole";
     ctx.cert_file = swoole::test::get_ssl_dir() + "/passwd.crt";
     ASSERT_TRUE(ctx.create());
+}
+
+TEST(ssl, failed_create) {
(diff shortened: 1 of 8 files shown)


NathanFreeman:
Thank you for your PR! However, this scenario is not common in actual development. Since set() must be called before start(), all SSL configurations can be fully set up in a single call, making multiple calls unnecessary. If configuration adjustments are truly needed, you can simply modify the first set() call directly.

binaryfire:
@NathanFreeman You’re right that repeated `set()` calls aren’t needed here. I’ve removed that handling. I’ve also added fixes for related SSL initialization failures found since the PR was opened: SNI-only listeners, invalid certificates, and failed context creation. Please take another look when you have a moment.

binaryfire:
Follow-up: This PR now also supports server TLS configurations that provide certificates only through SNI. It rejects invalid or incomplete certificate settings, and failed SSL context creation now clears the incomplete context so a later attempt starts cleanly. Tests cover SNI-only listeners, invalid default certificates, incomplete settings, and failed-context cleanup.

PR #6215, opened 2026-09-09
Reject oversized HTTP Upgrade values

binaryfire:
`Request::parse()` accepts caller-provided HTTP data without the server header-size check. Upgrade parsing copied the complete value into the current thread buffer and then appended a null byte. A value at least as large as that buffer wrote past its end and corrupted heap state.

This change returns false before copying a value that cannot fit together with the terminating byte. Parsing for valid Upgrade values is unchanged.

The regression passes a 128 KiB Upgrade value through `Request::parse()`. On unchanged 6.1 it prints `DONE`, then aborts with `double free or corruption (!prev)`; Valgrind also reports the invalid writes. The guarded implementation completes normally and under Valgrind while retaining the original header value.

### DIFF
--- a/ext-src/swoole_http_request.cc
+++ b/ext-src/swoole_http_request.cc
@@ -327,7 +327,7 @@ static void http_request_add_upload_file(HttpContext *ctx, const char *file, siz
 }
 
 bool swoole_http_token_list_contains_value(const char *at, size_t length, const char *value) {
-    if (0 == length) {
+    if (0 == length || length >= sw_tg_buffer()->size) {
         return false;
     }
     if (SW_STRCASEEQ(at, length, value)) {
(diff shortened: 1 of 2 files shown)


ext-src/swoole_http_request.cc
```diff
@@ -327,30 +327,28 @@ static void http_request_add_upload_file(HttpContext *ctx, const char *file, siz
 }
 
 bool swoole_http_token_list_contains_value(const char *at, size_t length, const char *value) {
-    if (0 == length) {
-        return false;
-    }
```
NathanFreeman, line 332: Thank you for your contribution, but I think this PR might be overly complex. This is essentially an invalid input — the Upgrade value exceeds the legal length limit. We could simply check whether its length exceeds the buffer size before parsing, and return false immediately if it does, without having to refactor the entire parsing logic.

binaryfire: I've updated this to keep the existing parsing logic and return false when the Upgrade value doesn't fit in the buffer. Thanks for the suggestion.

PR #6220, opened 2026-09-11
Preserve HTTP pipelined response order and receive pauses

binaryfire:
A pipelined HTTP/1 connection can dispatch a later request while an earlier handler is suspended. Both handlers send directly to the same connection, so the later response can arrive first.

This dispatches one dynamic HTTP/1 request at a time on each connection and resumes parsing buffered requests after the final response output is accepted. Completion uses the existing send metadata, without buffering or copying whole responses. It also sends `100 Continue` through the normal connection output path so it cannot overtake an earlier response.

`Server::pause()` could also be undone when pending output drained or queued-input backpressure cleared. Both transitions now leave reading disabled until `Server::resume()`.

The HTTP regression fails on unchanged 6.1 in both server modes. Two process-mode regressions cover both ways an application pause could be lost.

### DIFF
--- a/ext-src/php_swoole_http.h
+++ b/ext-src/php_swoole_http.h
@@ -162,6 +162,7 @@ struct Context {
     uchar parse_body : 1;
     uchar parse_files : 1;
     uchar http2 : 1;
+    uint16_t send_ext_flags;
 
     zval zsocket;
     uint32_t stream_id;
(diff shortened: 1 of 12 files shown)


binaryfire:
Follow-up: this now also keeps `Server::pause()` in effect when pending output drains or queued-input backpressure clears. Both paths previously restored reads before `Server::resume()`. Two process-mode regressions cover those cases.

PR #6231, opened 2026-09-12
Close duplicated standard IO handles after thread exit

binaryfire:
## Summary

Each PHP child thread opens its own `STDIN`, `STDOUT`, and `STDERR` streams. PHP duplicates the process descriptors for those streams, but Swoole marked the duplicates as non-closing during registration and request shutdown. Every completed thread therefore left three descriptors open. Thread-mode Server worker restarts followed the same path.

This change lets child request shutdown close its duplicated standard-I/O descriptors while preserving the main request's existing shutdown behavior.

## Testing

Added a Linux thread regression that compares the process descriptor count before starting a child thread and after joining it. It increases by three on the unchanged 6.1 branch and remains unchanged with this fix.

The existing thread standard-I/O test and representative user-thread and Server thread paths pass with the fix.

### DIFF
--- a/ext-src/php_swoole.cc
+++ b/ext-src/php_swoole.cc
@@ -1304,8 +1304,10 @@ PHP_RSHUTDOWN_FUNCTION(swoole) {
         stream->flags |= PHP_STREAM_FLAG_NO_CLOSE;
     };
     /* do not close the stdout and stderr */
-    php_swoole_set_stdio_no_close(ZEND_STRL("STDOUT"));
-    php_swoole_set_stdio_no_close(ZEND_STRL("STDERR"));
+    if (sw_is_main_thread()) {
+        php_swoole_set_stdio_no_close(ZEND_STRL("STDOUT"));
+        php_swoole_set_stdio_no_close(ZEND_STRL("STDERR"));
+    }
 #endif
 
     return SUCCESS;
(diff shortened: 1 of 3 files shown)


NathanFreeman:
Thanks for your work, but `false` cannot be passed in here—otherwise users could directly close STDIN, STDOUT, and STDERR from within PHP code. This bug has already been fixed in https://github.com/swoole/swoole-src/pull/6237.

matyhtf:
Thanks for the report and the patch - the leak is real and the analysis of the mechanism is correct. Each PHP child thread opens its own `php://stdin`/`php://stdout`/`php://stderr` streams, PHP duplicates the process descriptors for them, and Swoole marked those duplicates with `PHP_STREAM_FLAG_NO_CLOSE`, so they were never closed when the thread exited.

We cannot take the `thread_register_stdio_file_handles(false)` part of the change though, for the reason already raised in the review above: clearing the flag makes those streams closeable, which also lets user code running in a child thread close `STDIN`/`STDOUT`/`STDERR`. The descriptors have to be closed by Swoole itself instead.

That is the approach already merged on `master` (47b87cdb2, #6245, plus a follow-up) and it has now been backported to this branch:

- 6.1: b9d803750 `fix(thread): close the duplicated stdio file descriptors after thread exit`
- 6.2: 50bed42e7

It keeps the streams marked as `NO_CLOSE`, records the three descriptors at registration time and closes them right after `php_request_shutdown()`. It also ships the same kind of Linux regression test you proposed: after 8 threads have been created and joined, the process descriptor count must not change (it grows by 24 without the fix).

Closing this one as superseded - thanks again for the detailed write-up, it is what let us pin down the descriptor lifetime.

Issue #6243, opened 2026-09-19
[6.1.10] undeclared identifier 'sw_usleep'


BjarneDMat:
`swoole-6.1.10/src/core/misc.cc:174:9: error: use of undeclared identifier 'sw_usleep'; did you mean 'usleep'?`

I can see, you've completely rewritten `sw_atomic_futex_wait` between **6.1.9** & **6.1.10**.
In **6.1.9** you use `usleep` - not `sw_usleep`

```
Model Identifier: MacPro5,1     macOS 15.7.8 24G824 x86_64
Xcode 26.3 17C529               Command Line Tools 26.3.0.0.1.1771626560
```
you've got this : 
```
044 #ifdef HAVE_FUTEX
...
077 #elif defined(_WIN32)
...
153 #else
-> macOS ends here
186 #endif
```

NathanFreeman:
I'll take a look later and see what's going on.

PR #6244, opened 2026-09-20
Fix build on platforms without futex support

lazerg:
`sw_atomic_futex_wait()` calls `sw_usleep()` in the fallback branch used when neither futex nor Win32 is available. That name only exists on master, where `include/swoole_posix.h` maps it to `usleep`. The 6.1 branch has no such header, so since 6.1.10 the build fails on macOS and any other non-futex platform with "use of undeclared identifier 'sw_usleep'".

The call now uses `usleep()` directly, matching the rest of 6.1 (`src/core/base.cc`, `src/lock/barrier.cc`, `src/network/socket.cc`). Master is unaffected and should keep `sw_usleep`, since it needs the alias for the Windows shim. The 6.2 branch carries the same broken line and also has no `swoole_posix.h`, so it needs this fix as well, either through a merge from 6.1 or on its own.

Fixes #6243


lazerg:
These failures do not come from this change.

The ubuntu unit tests, the zts-thread jobs and Hyperf 8.1 are red on the current 6.1 head as well. The zts-thread jobs never reach the code, they stop at setup with "Invalid PHP version in php-version input: 8.1-zts".

The macOS job now gets past the build, which is the point of the patch. Two of the tests it reports, `swoole_atomic/timeout.phpt` and `swoole_atomic/wait_ex.phpt`, fail on master too, where the same fallback branch compiles because `sw_usleep` is defined there. `wakeup(n)` in that branch releases one waiter instead of n, so both tests hang on their remaining waiters. That is separate from this patch and I left it alone.

